Issue #27, opened 2016-12-29
Add CMake files


jupp0r:
Bazel is not very wide-spread yet, we should support CMake as a first-class citizen in order to make consuming this library easier.

gjasny:
Please have a look at #28.

PR #28, opened 2017-01-12
Add CMake build

gjasny:
Works on macOS with ``brew install cmake protobuf google-benchmark`` as a prerequisite.


gjasny:
In a follow-up PR we could embed the civetweb web server object files directly into the prometheus-cpp library via object libraries.

jupp0r:
Thanks! LGTM, I'll add this to CI and update the README. This fixes #27.

PR #34, opened 2017-01-23
Properly install public headers

jupp0r:
This change exposes public headers. In the future the amout of exposed headers could be reduced my using private implementation techniques.


gjasny:
Why don't you use `prometheus-cpp` as header prefix?

jupp0r:
I like the namespace matching the include directory.

include/prometheus/handler.h
```diff
@@ -3,10 +3,10 @@
 #include <memory>
 #include <vector>
 
-#include "CivetServer.h"
+#include <prometheus/collectable.h>
```
gjasny: Could you include your own headers with either `"prometheus/collectable.h"` or just a plain `"collectable.h"`.

lib/CMakeLists.txt
```diff
@@ -57,6 +45,7 @@ add_library(prometheus-cpp
 
 # TODO(gj) make all PRIVATE
 target_link_libraries(prometheus-cpp PUBLIC ${PROTOBUF_LIBRARIES})
+target_include_directories(prometheus-cpp PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../include> $<INSTALL_INTERFACE:include>)
```
gjasny: There is no need for the `$<INSTALL_INTERFACE>` genex here because the `INCLUDES DESTINATION ...` below does the same thing.

PR #35, opened 2017-02-03
Hide civetweb from public interface

gjasny:


lib/exposer.cc
```diff
@@ -4,16 +4,23 @@
...
-      metrics_handler_(collectables_, *exposer_registry_) {
+      metrics_handler_(
+          new detail::MetricsHandler{collectables_, *exposer_registry_}) {
   RegisterCollectable(exposer_registry_);
-  server_.addHandler("/metrics", &metrics_handler_);
+  server_->addHandler("/metrics", metrics_handler_.get());
+}
+
+Exposer::~Exposer() {
+  // needed to properly destroy server_
```
jupp0r: We should either call removeHandler() or reverse the order of the `server_` and metrics_handler_` members. The way it is right now, there's a race condition where `metrics_handler_` is destroyed and `server_` dispatches to its dangling pointer.

jupp0r:
A rebase to master should fix the build, apart from that, LGTM.

PR #36, opened 2017-02-03
Upgrade to protobuf 3.2.0 and generate metrics

gjasny:


jupp0r:
Nice, this is more elegant than what I came up with: https://github.com/jupp0r/prometheus-cpp/compare/feature/update-protobuf?expand=1

gjasny:
Seems you did not get my message that I'm working on this. It took me a while until I realized that I must invoke the `protoc` compiler from within the `prometheus_client_model` repository. The pesky protobuf compiler insists in prefixing the generated files with the path to the proto file. So I could not just reference the `metrics.proto` from `prometheus-cpp`. If it would not do that one could just fetch the proto file:

```
http_file(
    name = "client_model_metrics",
    url = "https://raw.githubusercontent.com/prometheus/client_model/e2da43ae71fe22f457da00bb0b1f4fcaec9113c2/metrics.proto",
    sha256 = "528b6acf6cb424f338672f375e333594a08515e5015b1d2f4d2445aa496320d0",
)
```

PR #39, opened 2017-03-06
Build fixes for various platforms

gjasny:


include/prometheus/counter_builder.h
```diff
@@ -19,8 +19,6 @@ detail::CounterBuilder BuildCounter();
 namespace detail {
 class CounterBuilder {
  public:
-  friend CounterBuilder BuildCounter();
```
jupp0r: Nice catch

lib/handler.cc
```diff
@@ -74,7 +74,7 @@ bool MetricsHandler::handleGet(CivetServer* server,
             "HTTP/1.1 200 OK\r\n"
             "Content-Type: %s\r\n",
             content_type.c_str());
-  mg_printf(conn, "Content-Length: %lu\r\n\r\n", body.size());
+  mg_printf(conn, "Content-Length: %lu\r\n\r\n", static_cast<unsigned long>(body.size()));
```
jupp0r: Did you try `mg_printf(conn, "Content-Length: %zu\r\n\r\n", body.size());` ?

PR #40, opened 2017-03-06
Fix distance calculation

gjasny:


lib/histogram.cc
```diff
@@ -10,10 +11,10 @@ Histogram::Histogram(const BucketBoundaries& buckets)
 
 void Histogram::Observe(double value) {
   // TODO: determine bucket list size at which binary search would be faster
-  auto bucket_index = std::max(
-      0L, std::find_if(bucket_boundaries_.begin(), bucket_boundaries_.end(),
-                       [value](double boundary) { return boundary > value; }) -
-              bucket_boundaries_.begin());
+  auto bucket_index = static_cast<std::size_t>(std::distance(
```
jupp0r: Note to myself: use `std::upper_bound` the next time.

Issue #51, opened 2017-04-17
Exposer implicit construction breaks readme examples


bahostetterlewis:
Currently the readme says to create an exposer as follows

```c++
// create an http server running on port 8080
  auto exposer = Exposer{"127.0.0.1:8080"};
```
However, the exposer implicitly constructs from the string, and tries to call the copy constructor, which fails.   
Potentially avoid this by making the Exposer constructor explicit.

Info:
using c++14 and gcc 6.2 on Ubuntu 14.04

PR #52, opened 2017-04-17
Exposer constructor marked explicit

odeits:
Prevents accidental conversion from std::string to Exposer

Relates to issue #51


jupp0r:
Thanks for finding and fixing this.

PR #74, opened 2017-09-27
Update bazel build module

mapx:
`prometheus-cpp` is not a legal workspace name, thus it's changed to `prometheus_cpp`, and so is the name of primary target.

Hope it's fine.


jupp0r:
Thanks!

Issue #75, opened 2017-10-30
Family::Add function should check for existing metric


mikea:
Without `Add` function doing existing labels check, it is very inconvenient to work with dynamic label values. I have to introduce a map myself and keep track of created counters. Instead it will be trivial for Family to check.

jupp0r:
Do you think throwing would be appropriate in that situation?

mfontanini:
I've had to do this in my code as well. The ideal solution would be if `Family::Add` checked if there's already a metric with those labels and if it does exist, then it returns the existing metric. Right now the code is very broken if you call `Add` with the same labels twice as you'll be returning a reference to an already deleted pointer (`unique_ptr` deletes it right away as the `insert` will fail due to the element already being there).

jupp0r:
That makes sense, I just wonder if it might be better to not silently ignore the double add, but trigger something that reveals it to the user.

mikea:
@jupp0r @mfontanini correctly handling is way, way better than revealing to the user. Why would you like to throw?

jupp0r:
I just thought that it might be nice to realize you are double-adding a metric, which you might consider a logic error in your code. I don't prefer throwing, I just wanted to include that as an option in the discussion we are having :) Personally, I actually prefer just returning the existing metric.

mikea:
> Personally, I actually prefer just returning the existing metric.

+100 to that. 

> which you might consider a logic error in your code

I can hardly imagine a situation when returning the existing metrics is not what use wants.

jupp0r:
Thanks for the input, returning the existing metric it is, then.

mikea:
PR: https://github.com/jupp0r/prometheus-cpp/pull/76

mikea:
this is fixed.

PR #76, opened 2017-11-03
supporting multiple adds with same labels. Fixes issue #75

mikea:


tests/family_test.cc
```diff
@@ -75,6 +75,13 @@ TEST_F(FamilyTest, Histogram) {
   EXPECT_THAT(collected[0].metric(0).histogram().sample_count(), Eq(1));
 }
 
+TEST_F(FamilyTest, add_twice) {
```
jupp0r: Thanks for writing a test for this

include/prometheus/family.h
```diff
@@ -67,14 +67,26 @@ T& Family<T>::Add(const std::map<std::string, std::string>& labels,
...
+
+  if (metrics_iter != metrics_.end()) {
+#ifndef NDEBUG
+    auto labels_iter = labels_.find(hash);
+    assert(labels_iter != labels_.end());
+    const auto &old_labels = labels_iter->second;
+    assert(labels == old_labels);
+#endif
+    return *metrics_iter->second;
+  } else {
```
jupp0r: Minor nitpick: you could leave out the else because you already returned at the end of the last block

PR #78, opened 2017-11-29
Add Set method to counter

ggarber:


ggarber:
I have a counter internally (maintaining the number of connections) and I would like to expose its value periodically as a counter metric.     

I thought it could be useful but if it is not included in other client libraries I agree it is better not to include it.    I can expose my metric as a gauge or try to refactor my code to use a prometheus counter instead of my internal counter.

Feel free to close it.

jupp0r:
The number of active connections would better be modelled by a gauge. Alternatively, you could create two counters - one for opened connections and one for closed connections. Then you could query the total connections and the active connections separately.

mikea:
I agree that counter shouldn't have Set method (you'll be able to decrease it). To synchronize two counters you can always do `counter.Increment(otherCounter.Value() - counter.Value());`

Issue #77, opened 2017-11-29
Make serializers public


ggarber:
Make json/text/protobuf... serializers public so that we can reuse them without using the Exposer.   Use case is to add a /metrics method to an existing HTTP server.

jupp0r:
Makes sense to share that code.

PR #79, opened 2017-12-05
Exposing metrics serializers

mikea:
This is useful for exporting metrics over different transports
(e.g. writing to file for batch jobs)

Fixes issue #77


jupp0r:
Thanks for the contribution! Can you also move the headers in the bazel `BUILD` file the same way you did in `CMakeLists.txt`? That should fix the travis build.

mikea:
@jupp0r done

PR #80, opened 2017-12-14
Histogram boundary value should be inclusive

khabinov:
Boundary value should be inclusive to make it consistent with other `Prometheus` clients:

* `Java` https://github.com/prometheus/client_java/blob/fb4161a2be99074f1d2fd254238193afa4aa0398/simpleclient/src/main/java/io/prometheus/client/Histogram.java#L238

* `Golang` https://github.com/prometheus/client_golang/blob/661e31bf844dfca9aeba15f27ea8aa0d485ad212/prometheus/histogram.go#L250 where
```go
func SearchFloat64s(a []float64, x float64) int {
    return Search(len(a), func(i int) bool { return a[i] >= x })
}
```

* `Python` https://github.com/prometheus/client_python/blob/24d4d95d3fee82e1b5d8679b69b340dab5815eed/prometheus_client/core.py#L891

* `Ruby` https://github.com/prometheus/client_ruby/blob/726512562b354995ab4d325bc4f654833eb225e5/lib/prometheus/client/histogram.rb#L29

### DIFF
--- a/lib/histogram.cc
+++ b/lib/histogram.cc
@@ -17,7 +17,7 @@ void Histogram::Observe(double value) {
   auto bucket_index = static_cast<std::size_t>(std::distance(
       bucket_boundaries_.begin(),
       std::find_if(bucket_boundaries_.begin(), bucket_boundaries_.end(),
-                   [value](double boundary) { return boundary > value; })));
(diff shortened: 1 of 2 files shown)


jupp0r:
Thanks!

PR #81, opened 2017-12-20
No need to traverse twice

parsifal-47:


jupp0r:
Good catch, I did not see that this is redundant now that we correctly expose cumulative histograms.

PR #82, opened 2018-01-03
Fix build compatibility with bazel 0.9.0

th0br0:
bazel 0.9.0 no longer accepts abridged/md5 hashes.


jupp0r:
Thanks for the contribution!

PR #83, opened 2018-01-08
split civetweb into c and c++ libraries

Foran:
To enable projects that include prometheus-cpp to optionally exclude the civetweb cpp wrapper library as a direct dependency.

In my case I'm building a web-app based on civetweb, but don't want to link with the c++ wrapper.

### DIFF
--- a/repositories.bzl
+++ b/repositories.bzl
@@ -16,13 +16,11 @@ _CIVETWEB_BUILD_FILE = """
 licenses(["notice"])  # MIT license
 
 cc_library(
-    name = "civetweb",
+    name = "libcivetweb",
     srcs = [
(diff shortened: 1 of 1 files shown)


jupp0r:
Thanks for the contribution, does that mean that you are implementing your own `Exposer` on top of your existing `civetweb` server? If so, you could also not link against the C part of civetweb, right?

Foran:
You're welcome :-)  I have a light weight C service that uses civetweb (so that in can run in resource constrained places) and can optionally dynamically link a module that exports to prometheus.  Ideally I'd like to use prometheus-cpp.   If I try to link a seperate copy of civetweb, it fails due to duplicate symbols.  Make sense?

PR #84, opened 2018-01-09
the civet server should be available externally of the exposer

jb--:
I would like to have the Civet server available outside of the Exposer class. The reasoning for this is that I would like to add a health check endpoint to my application and I do not want to start a separate web server.


jupp0r:
I don't think this is the best way to solve your problem. `Exposer` was added to have a no-setup solution for exposing metrics. For more custom things, you should rather register `Collectables` with you own http server infrastructure than use the one held by `prometheus-cpp`.

I'm having concerns about exposing Civetweb on the public API, because I consider it an implementation detail of the library, that you might choose to use or not.

Makes sense?

PR #85, opened 2018-01-21
WIP: implement summary metric

Daejiv:
**Issue:** The summary metric type is not implemented, see #4

This implements the summary type. It uses the same implementation as the [java client](https://github.com/prometheus/client_java)

For not it's still work in progess, since I'll want to do some testing in our dev/prod environment and maybe find some performance improvements.

### DIFF
--- a/include/prometheus/summary.h
+++ b/include/prometheus/summary.h
@@ -1,0 +1,103 @@
+#pragma once
+
+#include <atomic>
+#include <chrono>
+#include <list>
+#include <mutex>
+#include <vector>
+
+#include "metrics.pb.h"
+
+#include "prometheus/metric.h"
+
+namespace prometheus {
+
+namespace detail {
+class CKMSQuantiles {
+ public:
+  struct Quantile {
+    const double quantile;
+    const double error;
+    const double u;
+    const double v;
+
+    Quantile(double quantile, double error);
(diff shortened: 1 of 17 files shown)


lib/registry.cc
```diff
@@ -29,6 +29,15 @@ Family<Histogram>& Registry::AddHistogram(
   return *histogram_family;
 }
 
+Family<Summary>& Registry::AddSummary(
+    const std::string& name, const std::string& help,
+    const std::map<std::string, std::string>& labels) {
+  std::lock_guard<std::mutex> lock{mutex_};
+  auto histogram_family = new Family<Summary>(name, help, labels);
```
jupp0r: this should be called `summary_family`

include/prometheus/summary.h
```diff
@@ -0,0 +1,103 @@
...
+    const double error;
+    const double u;
+    const double v;
+
+    Quantile(double quantile, double error);
+  };
+
+ private:
+  struct Item {
+    /*const*/ double value;
```
jupp0r: What prevents these fields from being const?

lib/registry.cc
Daejiv: Copy&paste error, will fix later.

PR #86, opened 2018-01-21
fix compile for protobuf 2.x

Daejiv:
**Issue:** Due to dependency problems on certain systems using protobuf 3.x is not an option.

This fixes any compile issues when building with protobuf 2.x.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,4 @@
 bazel-*
+
+/.idea/
+/cmake-build-*/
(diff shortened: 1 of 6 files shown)


jupp0r:
I think I'd rather remove json support altogether. It's never been supported by prometheus anyway (for good reason), so I'll rather remove it than maintain it.

jupp0r:
Closing in favor of #93

PR #87, opened 2018-02-20
Fix link issue with ZLib

Thomas-Barbier-1A:
This is small fix for issue I had when compiling (in my slightly particular environment): linking with old system ZLib instead of the one I wanted.

libprotobuf.so: undefined reference to `deflate'
libprotobuf.so: undefined reference to `deflateEnd'
libprotobuf.so: undefined reference to `inflate'
libprotobuf.so: undefined reference to `inflateInit2_'
libprotobuf.so: undefined reference to `inflateEnd'

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,7 @@ endif()
 
 find_package(GoogleBenchmark)
 find_package(Protobuf REQUIRED)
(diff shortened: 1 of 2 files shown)


jupp0r:
I'll need to add bazel support as well.

PR #88, opened 2018-02-22
Switch bazel WORKSPACE instructions from git_repository to http_archive.

isturdy:
git_repository is deprecated--Bazel recommends http_archive where possible, as it is faster and better supported internally (https://docs.bazel.build/versions/master/be/workspace.html#git_repository).


jupp0r:
Thanks for the fix @isturdy

PR #92, opened 2018-02-25
Prevent usage of broken regex implementation

### LINKED ISSUE #66: Old versions of g++/clang crash on name_checker regexes
This is not a bug in the project perse but since it does impact the usability I thought I'd file an issue. 

I am building with a relatively ancient g++ (4.8.4) and while it compiles fine when it runs and tries to compile the regex it throws an exception.

Specifying a different regex parser makes the problem go away, I'll submit a PR with that if you are okay accepting that change.

gjasny:
GNU libstdc++ prior to GCC 4.9 immediately throws.

Visual Studio 2013 messes up static initialization.

Closes #66

### DIFF
--- a/lib/check_names.cc
+++ b/lib/check_names.cc
@@ -2,21 +2,35 @@
 
 #include <prometheus/check_names.h>
 
+#if defined(__GLIBCXX__) && __GLIBCXX__ <= 20150623
+#define STD_REGEX_IS_BROKEN
+#endif
+#if defined(_MSC_VER) && _MSC_VER < 1900
+#define STD_REGEX_IS_BROKEN
+#endif
+
 namespace prometheus {
 bool CheckMetricName(const std::string& name) {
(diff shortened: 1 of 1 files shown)


lib/check_names.cc
```diff
@@ -2,21 +2,35 @@
 
 #include <prometheus/check_names.h>
 
+#if defined(__GLIBCXX__) && __GLIBCXX__ <= 20150623
```
gjasny: Unfortunately libstdc++ is not properly versioned. The date jumps back and forth for bugfix releases: [GCC releases](https://gcc.gnu.org/develop.html#timeline). I picked the value from latest GCC 4.8 on centos7.

jupp0r:
Theoretically we could move the statically initialized regex out of the function scope and make the check work with VS2013. I don't think it's worth the hassle, though. Thanks for the contribution!

PR #94, opened 2018-02-26
asan: define macro and don't invalidate bazel's cache.

g-easy:


g-easy:
The define may not be strictly needed for prometheus-cpp (sorry), it's definitely used in abseil and e.g. chromium code. I'm just trying to keep the convention going. :)

As for the `cc_output_directory_tag`, I've found that helps a lot with incremental builds.

jupp0r:
Both makes sense, thanks for the contrib! (I was googling `#define ADDRESS_SANITIZER` and found the chromium references).

PR #100, opened 2018-03-20
Add cmake option to build shared lib (.so file in nix)

ganibc:


gjasny:
Hi, can't you just pass `-DBUILD_SHARED_LIBS` to tell CMake to make it a shared library?
In case you want to avoid the extra shared lib, you could also tell CMake to enable position independent code.

ganibc:
Hi, I'm not use to cmake. I will try this tomorrow. I will close this pull request first.

PR #106, opened 2018-04-20
Do not link rt on Apple

gjasny:


gjasny:
Hello @drigz @isturdy @mapx!

Sorry to bother you, but could you please have a look at the macOS bazel build? civetweb should not   link to `rt` on MacOS. Right now I feel lost somehow because I don't know how to do it _properly_ with Bazel.

Thanks,
Gregor

drigz:
Hey @gjasny! I would add config settings: ([example](https://github.com/bazelbuild/bazel/blob/96d3c91c714544584c9174759bedebf2a6be5e71/src/conditions/BUILD#L19))

```
config_setting(
    name = "darwin",
    values = {"cpu": "darwin"},)

config_setting(
    name = "darwin_x86_64",
    values = {"cpu": "darwin_x86_64"},
)
```

and then use `select()` to adjust the linkopts: ([example](https://github.com/bazelbuild/bazel/blob/96d3c91c714544584c9174759bedebf2a6be5e71/src/BUILD#L10))

```
    linkopts = [
        "-lpthread",
    ] + select({
        ":darwin": [],
        ":darwin_x86_64": [],
        "//conditions:default": ["-lrt"],
    }),
```

Will you add it to this PR?

repositories.bzl
```diff
@@ -1,6 +1,15 @@
 _CIVETWEB_BUILD_FILE = """
 licenses(["notice"])  # MIT license
 
+config_setting(
```
gjasny: @drigz Is that the proper place for config settings? Would root `BUILD` file be the better place? If yes, how could I reference them?

drigz: IMO it's great as it is. Since they're only used in one place, it's clearer to declare them where they're used.

For future reference, if they were in the root BUILD file you could refer to them with @com_github_jupp0r_prometheus_cpp//:darwin.

PR #110, opened 2018-04-23
fix exposer and server threads synhnoniztion

Lokimed:


include/prometheus/exposer.h
```diff
@@ -25,7 +26,12 @@ class Exposer {
 
  private:
   std::unique_ptr<CivetServer> server_;
-  std::vector<std::weak_ptr<Collectable>> collectables_;
+  // \todo maybe better create here class with mutex?
+  //       maybe best solution will be lock free list
+  //       But in any case container need syncronization defence
+  //       One thread can do RegisterCollectable and another thread in server can do collect from container
+  std::mutex collectables_guard_;
+  std::list<std::weak_ptr<Collectable>> collectables_;
```
jupp0r: Is there any reason you changed the container from a vector to a list?

include/prometheus/exposer.h
```diff
@@ -25,7 +26,12 @@ class Exposer {
 
  private:
   std::unique_ptr<CivetServer> server_;
-  std::vector<std::weak_ptr<Collectable>> collectables_;
+  // \todo maybe better create here class with mutex?
+  //       maybe best solution will be lock free list
+  //       But in any case container need syncronization defence
+  //       One thread can do RegisterCollectable and another thread in server can do collect from container
+  std::mutex collectables_guard_;
```
jupp0r: A better way to solve this would be to have exposer take a container of collectables in its constructor. This would alleviate the need for locks. The list of collectables should be finite anyway, I don't think we need the dynamics of the `RegisterCollectable()` function.

Lokimed: Okay. You are right. I will do fix.

include/prometheus/exposer.h
Lokimed: Reason was delete element when weak_ptr.lock return null shared_pointer, but i see below comment. If exposer will register container only in constructor  so vector is good. I will do fix and will change commit.

Lokimed:
I hope now this look good

include/prometheus/exposer.h
```diff
@@ -18,7 +18,10 @@ class MetricsHandler;
 
 class Exposer {
  public:
-  explicit Exposer(const std::string& bind_address,
+  using CollectableContainer = std::vector<std::weak_ptr<Collectable>>;
+
+  explicit Exposer(CollectableContainer&& collectables,
```
jovan-wong: Remove `explicit` because of the 2 params without `default-arguments` here.

include/prometheus/exposer.h
```diff
@@ -18,7 +18,10 @@ class MetricsHandler;
...
+                   const std::string& bind_address,
```
jovan-wong: `bind` is a verb, so use `bound_address` or `address` instead of it?

lib/exposer.cc
```diff
@@ -10,14 +10,17 @@
 
 namespace prometheus {
 
-Exposer::Exposer(const std::string& bind_address, const std::string& uri)
+Exposer::Exposer(CollectableContainer&& collectables,
```
jovan-wong: `CollectableContainer&& container`

tests/integration/sample_server.cc
```diff
@@ -29,8 +26,12 @@ int main(int argc, char** argv) {
   auto& second_counter = counter_family.Add(
       {{"another_label", "value"}, {"yet_another_label", "value"}});
 
-  // ask the exposer to scrape the registry on incoming scrapes
-  exposer.RegisterCollectable(registry);
+  // create container which must contain all registry which will be scraped
```
jovan-wong: `all registry` --> `all registries`

include/prometheus/exposer.h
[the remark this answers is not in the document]
KholdStare: The code style in the rest of the code base so far is having references attached to types, rather than variable names @jovany-wang . Why do you suggest changing it?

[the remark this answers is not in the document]
jovany-wang: Oops, sorry. That's my personal prefer. 
It's fine to follow the code style of  the codebase on master branch.

Issue #101, opened 2018-03-29
Refactor into smaller, combinable units


benjamin-bader:
I'm interested in using portions of this library, specifically the various metrics implementations.  I do not wish to take dependencies on protobuf or civet.  I think there's a lot of value in having Counter, Histogram, etc available independently of transport concerns.

It seems reasonable that we might refactor the library into separate components, that would still be combined into one binary but which could also be incorporated individually - for example, in pseudo-cmake:

```cmake
add_library(metrics, ${METRICS_SRCS})
add_library(metrics_protobuf ${PROMETHEUS_PROTO_SRCS})
add_library(civet_exposer, ${EXPOSER_SRCS})
add_library(prometheus-cpp
  $<TARGET_OBJECTS:metrics>
  $<TARGET_OBJECTS:metrics_protobuf>
  $<TARGET_OBJECTS:civet_exposer>
)

# etc
```

I haven't looked too deeply into the feasibility, but (as one unfamiliar with the code) it doesn't seem too far-fetched.  Is there any interest in supporting this kind of refactor?

jupp0r:
Metrics currently all implement `Collectable`, which depends on protobuf representations of prometheus' metrics. With protobuf content-type support gone in prometheus 2.0, we could move away from using protobuf as a dependency alltogether. I'd definitely support that change.

Once this refactoring would be complete, splitting metrics out of the library would be easy.

gjasny:
I'll try getting rid of the protobuf dependency.

gjasny:
PR #105 should get rid of protobuf.

gjasny:
Please give #127 a try.

gjasny:
Hope current master is providing what you need.

Issue #23, opened 2016-11-11
Add Support for Pushgateway


olitheolix:
It would be great if the library had support for pushing to a gateway server, like the Python client (https://github.com/prometheus/client_python#exporting-to-a-pushgateway).

gjasny:
Could you please take a look at the current master and check if the push functionality works for you?

PR #112, opened 2018-06-13
Extract the `prometheus-core` library

flier:
related to #101, then we can use `prometheus-core` library without `civetweb`, and we may add #23 or other features later.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -1,14 +1,11 @@
 cc_library(
-    name = "prometheus_cpp",
+    name = "prometheus_core",
     srcs = [
         "lib/check_names.cc",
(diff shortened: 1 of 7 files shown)


flier:
@gjasny I think we can add more library base on the `prometheus-core`, for example #113 may be named `prometheus-push` which depend on `curl` etc, and `prometheus-cpp` may become `prometheus-pull` base on `civetweb`. The user can choose which model they like and reduce the unnecessary dependency.

jupp0r:
@flier You could do the same based on a static library, right?

gjasny:
Closing in favor of #123

PR #113, opened 2018-06-13
push metrics to pushgateway

### LINKED ISSUE #23: Add Support for Pushgateway

flier:
resolve #23

### DIFF
--- a/include/prometheus/gateway.h
+++ b/include/prometheus/gateway.h
@@ -1,0 +1,59 @@
+#pragma once
+
+#include <memory>
+#include <sstream>
+
+#include <cpr/cpr.h>
+
+#include "histogram.h"
+#include "registry.h"
+
+namespace prometheus {
+
+class Gateway {
+ public:
+  typedef std::map<std::string, std::string> labels_t;
+
+  explicit Gateway(const std::string& uri, const std::string jobname,
+                   const labels_t* labels = nullptr,
+                   const std::string username = "",
+                   const std::string password = "");
+  ~Gateway() {}
+
+  void RegisterCollectable(const std::weak_ptr<Collectable>& collectable,
+                           const labels_t* labels = nullptr);
+
+  static const labels_t instance_label(void);
+
+  // Push metrics to the given pushgateway.
+  int Push(void) { return push(false); }
+
+  std::future<int> AsyncPush(void) { return async_push(false); }
+
+  // PushAdd metrics to the given pushgateway.
+  int PushAdd(void) { return push(true); }
+
+  std::future<int> AsyncPushAdd(void) { return async_push(true); }
+
+  // Delete metrics from the given pushgateway.
+  int Delete(void);
+
+  // Delete metrics from the given pushgateway.
+  cpr::AsyncResponse AsyncDelete(void);
+
+ private:
+  std::vector<std::pair<std::weak_ptr<Collectable>, std::string>> collectables_;
+  std::string uri_;
+  std::string jobname_;
+  std::string labels_;
+  std::string username_;
+  std::string password_;
+
+  std::stringstream job_uri(void) const;
+
+  int push(bool replacing);
+
+  std::future<int> async_push(bool replacing);
+};
+
+}  // namespace prometheus
--- a/lib/gateway.cc
+++ b/lib/gateway.cc
@@ -1,0 +1,149 @@
+#include <sys/param.h>
+#include <unistd.h>
+
+#include "prometheus/client_metric.h"
+#include "prometheus/gateway.h"
+#include "prometheus/serializer.h"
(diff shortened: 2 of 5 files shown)


lib/gateway.cc
```diff
@@ -0,0 +1,159 @@
...
+
+    labels_ = ss.str();
+  }
+}
+
+const Gateway::labels_t Gateway::instance_label(void) {
+  char hostname[MAXHOSTNAMELEN] = {0};
+
+  if (gethostname(hostname, MAXHOSTNAMELEN - 1)) {
+    hostname[0] = 0;
```
jupp0r: I'd rather return an empty label set here

include/prometheus/gateway.h
```diff
@@ -0,0 +1,59 @@
...
+
+  std::future<int> AsyncPush(void) { return async_push(false); }
+
+  // PushAdd metrics to the given pushgateway.
+  int PushAdd(void) { return push(true); }
+
+  std::future<int> AsyncPushAdd(void) { return async_push(true); }
+
+  // Delete metrics from the given pushgateway.
+  int Delete(void);
```
jupp0r: Please remove the void, it doesn't make any difference in C++ (as opposed to C).

lib/gateway.cc
```diff
@@ -0,0 +1,159 @@
...
+  if (labels) {
+    for (auto& label : *labels) {
+      ss << "/" << label.first << "/" << label.second;
+    }
+  }
+
+  collectables_.push_back(std::make_pair(collectable, ss.str()));
+}
+
+std::stringstream Gateway::job_uri(void) const {
```
jupp0r: why not return a string on this method?

lib/gateway.cc
```diff
@@ -0,0 +1,159 @@
...
+    futures.push_back(std::async(std::launch::async, [&] {
+      return replacing ? session.Post() : session.Put();
+    }));
+  }
+
+  return std::async(std::launch::async, [&] {
+    for (auto& future : futures) {
+      auto res = future.get();
+
+      if (res.status_code > 400) {
```
jupp0r: You are missing the 400 (bad request) error here, should be `>=`

tests/integration/sample_server.cc
```diff
@@ -13,6 +14,11 @@ int main(int argc, char** argv) {
   // create an http server running on port 8080
   Exposer exposer{"127.0.0.1:8080"};
 
+  // create a push gateway
+  auto labels = Gateway::instance_label();
+
+  Gateway gateway{"127.0.0.1:9091", "sample_server", &labels};
```
jupp0r: this will break the integration test run in CI, unless you also deploy a push gateway for the test.

gjasny:
Thanks for your contribution. I cherry-picked you changes on top of #123 and filed #124 for review.

include/prometheus/gateway.h
```diff
@@ -0,0 +1,59 @@
...
+#include <cpr/cpr.h>
+
+#include "histogram.h"
+#include "registry.h"
+
+namespace prometheus {
+
+class Gateway {
+ public:
+  typedef std::map<std::string, std::string> labels_t;
```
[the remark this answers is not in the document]
gjasny: fixed in #124

lib/gateway.cc
```diff
@@ -0,0 +1,159 @@
...
+#include "prometheus/gateway.h"
+#include "prometheus/serializer.h"
+#include "prometheus/text_serializer.h"
+
+namespace prometheus {
+
+const char* CONTENT_TYPE = "text/plain; version=0.0.4; charset=utf-8";
+
+Gateway::Gateway(const std::string& uri, const std::string jobname,
+                 const labels_t* labels, const std::string username,
```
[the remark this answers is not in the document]
gjasny: fixed in #124

lib/gateway.cc
gjasny: fixed in #124

include/prometheus/gateway.h
gjasny: fixed in #124

lib/gateway.cc
gjasny: I now calculate the value in the ctor. See #124

lib/gateway.cc
gjasny: fixed in #124

tests/integration/sample_server.cc
gjasny: I split it into pull mode `sample_server`(as it was before) and push mode `sample_client`

PR #114, opened 2018-06-22
added build options BUILD_TESTING and BUILD_CIVETWEB

bvgastel:
These options are to build the testing and to embed the webserver in prometheus-cpp. This way, the project can be used in a more flexible way. The options defaults to on.

Only adjusted the CMake file so it works for us.

Related to https://github.com/jupp0r/prometheus-cpp/pull/112 .

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,9 +10,9 @@ if(NOT DEFINED CMAKE_CXX_STANDARD AND UNIX)
 	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
 endif()
 
-OPTION(BUILD_CIVETWEB "Embed Civet webserver in prometheus" ON)
-ENABLE_TESTING()
-INCLUDE(CTest)
+option(BUILD_CIVETWEB "Embed Civet webserver in prometheus" ON)
+enable_testing()
+include(CTest)
 
 set(CMAKE_THREAD_PREFER_PTHREAD TRUE)
 find_package(Threads)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,7 +23,6 @@ if(APPLE)
 endif()
 
 if (BUILD_CIVETWEB)
-	find_package(ZLIB REQUIRED)
 	# civetweb
 	set(CIVETWEB_INCLUDE_DIR ${PROJECT_SOURCE_DIR}/3rdparty/civetweb/include)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,7 +7,7 @@ include(GNUInstallDirs)
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_CURRENT_LIST_DIR}/cmake)
 
 if(NOT DEFINED CMAKE_CXX_STANDARD AND UNIX)
-	set(CMAKE_CXX_STANDARD 11)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -1,76 +1,80 @@
...
+
 enable_testing()
-include(GNUInstallDirs)
 
+include(GNUInstallDirs)
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_CURRENT_LIST_DIR}/cmake)
 
 if(NOT DEFINED CMAKE_CXX_STANDARD AND UNIX)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+	set(CMAKE_CXX_STANDARD 11)
```
gjasny: Setting `CMAKE_CXX_STANDARD` won't work with CMake 2.8.12. Please revert

CMakeLists.txt
```diff
@@ -1,76 +1,80 @@
...
 endif()
 
+OPTION(BUILD_CIVETWEB "Embed Civet webserver in prometheus" ON)
```
gjasny: please use lowercase command names. Lowercase is also what CMake itself uses nowadays.

lib/CMakeLists.txt
```diff
@@ -1,25 +1,32 @@
-add_library(prometheus-cpp
+set(PROMETHEUS_SOURCES
```
gjasny: Instead of dynamically adding sources I'd like to see the lib split into two like #112 tries to do. Maybe we also should separate the files into separate directories. @jupp0r what do you think?

CMakeLists.txt
bvgastel: Oh, oops. This was by accident when I was fooling around. I will revert it.

lib/CMakeLists.txt
bvgastel: I considered grouping it in a core library a separate change, as work in #112 was already underway. I can do that for CMake, however, I'm not familiar with the other build tool.

CMakeLists.txt
bvgastel: Fixed.

lib/CMakeLists.txt
bvgastel: For us to integrate properly into our project, the options to disable building CIVETWEB and testing is important, as we do not use it and it gives us troubles building the project.

jupp0r: I think it would be nicer to be able to choose between a core library and one that comes with an embedded http server. My main concern is that an option will cause problems with distro packaging because only one option combination can ever be selected for a binary package. The core library option would allow us to distribute two packages so consumers can still use whatever option they want without building from source.

bvgastel: So in practice, you prefer the repository to be split into two parts, each one with a separate CMakeLists.txt and the core one is included as an external project in the one with the webserver? Otherwise running cmake could result in unresolved dependencies (as would be for us).

Do you prefer a directory layout for this? "core" and "webserver" directories?

gjasny: I filed #123 and #124. Once they are merged I'll add options to select what components to build.

gjasny:
Please give #127 a try.

gjasny:
Hope current master is providing what you need.

bvgastel:
It works, thanks!

PR #120, opened 2018-07-03
Fix compilation with Bazel on Windows

gjasny:


jupp0r:
Nice, I just noticed yesterday that the google benchmark guys closed my add a BUILD file issue, but I couldn't remember what I opened it for. Now I know, should have linked the issues.

PR #123, opened 2018-07-04
Split library into core and pull

gjasny:
I will work later on making parts of the build optional.

PR #124, opened 2018-07-04
Add push library

gjasny:
This is based on #113 (and on #123). I hope I addressed all open issues. I moved the hostname retrieval outside of the library because it is platform dependent code and one might want to customize it. On Windows one also has to do the "Winsock init dance".

What I'm unsure about is the selection of cpr as a http library. It's not available in Ubuntu or Homebrew. Maybe adding it as submodule like it was done for civetweb is the way to go. But CMake support will follow in another PR.


gjasny:
🛑 Do not merge without #123 and a rebase.

PR #142, opened 2018-09-04
Fix summary text format according to specs.

serialx:
According to the [Prometheus text exposition format][1] each quantile
named x is given as a separate line with the same name x and a label
{quantile="y"}.
But current Prometheus C++ implementation serializes summary named x
with a name x + "_quantile" for each quantile. This is off spec.

[1]: https://github.com/prometheus/docs/blob/master/content/docs/instrumenting/exposition_formats.md


jupp0r:
Thanks, good catch!

PR #146, opened 2018-09-13
Update the link to a working example of a server

0mp:


gjasny:
Thanks for spotting and fixing!

Issue #155, opened 2018-09-23
cannot build from tar.gz


skarlsson:
When I do a'source build without checking out a repo the build fails due to a hardcoded search path to prometheus-cpp/3rdparty/civetweb/include

I've already built all dependencys and placed them at /usr/local/include (and lib)

My build look like this:

wget -O prometheus-cpp.tar.gz "https://github.com/jupp0r/prometheus-cpp/archive/v0.4.2.tar.gz" && \
mkdir -p prometheus-cpp && \
tar \
  --extract \
  --file prometheus-cpp.tar.gz \
  --directory prometheus-cpp \
  --strip-components 1 && \
cd prometheus-cpp && \
mkdir build && cd build && \
cmake  -DCMAKE_BUILD_TYPE=Release ..

error...
CMake Error at cmake/civetweb-config.cmake:6 (message):
  File or directory /home/saka/prometheus-cpp/3rdparty/civetweb/include
  referenced by variable CIVETWEB_INCLUDE_DIR does not exist !
Call Stack (most recent call first):
  cmake/civetweb-config.cmake:10 (set_and_check)
  pull/CMakeLists.txt:1 (find_package)

gjasny:
Right now the CMake build embeds civetweb from a submodule directly into the pull library. If you don't use civetweb somewhere else in your process I'd suggest that you put civetweb into the expected directory.

I need to work on a patch to use an external civetweb library. But it might take some days.

skarlsson:
my problem is that I'm using somewhere else - that's why I need to build it out of source to be sure of the versions. copy from /usr/local/include will work for now. Thanks

gjasny:
Hi, could you please give https://github.com/jupp0r/prometheus-cpp/tree/external-dependencies a try? It's a rough version that allows one to use external civetweb (no cpr yet).
Please build with: `-DUSE_THIRDPARTY_LIBRARIES=OFF -DENABLE_PUSH=OFF -DENABLE_TESTING=OFF`

gjasny:
Could you please give #159 a try?

skarlsson:
Sorry missed this, will try tomorrow

skarlsson:
Thanks, works like a charm - can I which for cpr also?

yaminisridaran:
[...]

PR #159, opened 2018-09-25
Use system libraries instead of embedded ones

gjasny:
Closes: #155


.travis.yml
```diff
@@ -31,13 +31,54 @@ script:
   - bazel test --test_output=all //core/... //pull/...
   - bazel test --test_output=all //pull/tests/integration:scrape-test
   - bazel run -c opt //core/tests/benchmark
-  - mkdir _build
-  - cd _build
-  - cmake ..
+
+  - mkdir _build_internal_deps
+  - cd _build_internal_deps
```
jupp0r: pushd here as well?

.travis.yml
```diff
@@ -31,13 +31,54 @@ script:
   - bazel test --test_output=all //core/... //pull/...
   - bazel test --test_output=all //pull/tests/integration:scrape-test
   - bazel run -c opt //core/tests/benchmark
-  - mkdir _build
```
jupp0r: should we keep building with both internal and external deps?

gjasny: That makes it pretty convenient for users to embed civetweb for example. I also think it does not clutter the CMakeLists.txt files too much. For now I'd say we should keep them. Maybe flip the default to use external ones.

gjasny:
@jupp0r I believe it's ready to be merged. I did some small cleanups like moving the 3rd party deps build for travis into a separate file. Could you please look over the PR and merge if good?

PR #165, opened 2018-10-15
Refactor includes

jerryct:


gjasny:
Hi, I'm curious why you chose to use `#include <...>` over `#include "..."`?

jerryct:
> Hi, I'm curious why you chose to use `#include <...>` over `#include "..."`?
Most compilers implement <> in a way that you have to provide the include path to find the header file (although not required, because it is implementation defined: https://en.cppreference.com/w/cpp/preprocessor/include). In my projects I like to have control over the include path so using <> gives me a compile error if I missed a path.

Given that it is implementation defined where "" vs <> searches the header files, I think one should not bother too much whether to choose "" or <>. In my world everything is <>, which is a simple rule :)

jupp0r:
We use the [Google C++ Style](https://google.github.io/styleguide/cppguide.html) for this project, mainly because it's well documented and easy to point people to. This means that you should use `#include "prometheus/foo.h"` for project-internal includes.

PR #167, opened 2018-10-16
Document metric classes

jerryct:


jerryct:
Or clang-doc (https://clang.llvm.org/extra/clang-doc.html) - at one point in time.

gjasny:
please rebase to fix TravisCI

jerryct:
I rebased it.

core/include/prometheus/histogram.h
```diff
@@ -6,21 +6,52 @@
...
+  /// The bucket boundaries cannot be changed once the histogram is created.
+  Histogram(const BucketBoundaries&);
+
+  /// \brief Observe the given amount.
+  ///
+  /// The given amount selects the 'observed' bucket. The observed bucket is
+  /// chosen for which the given amount falls into the half-open interval [b_n,
+  /// b_n+1). The counter of the observed bucket is incremented. Also the total
+  /// sum of all observations is incremented.
+  void Observe(double);
```
gjasny: Is there a reason behind removing the argument names?

jerryct: It is the style currently used in many places of the library in the header file. Personally I would add the argument names also in the header (it also eases documenting with doxygen). I reverted the changes made to the argument names.

gjasny:
Do you have an idea where we could publish the generated doxygen files easily? I played around with "Read The Docs" [here](https://prometheus-cpp.readthedocs.io/en/read-the-docs/) and Breathe / Sphinx, but it does not look promising. Uploading to github pages might also be an option, but maybe you know something easier to set up?

gjasny:
Could you please rebase?

jerryct:
I rebased it.

jerryct:
Regarding documentation I have no better idea. I think github pages is the most straightforward way. Travis simply generates the html and uploads it to a special branch. There are some examples:

https://gist.github.com/francesco-romano/351a6ae457860c14ee7e907f2b0fc1a5

Why don't you like the read the docs output? We need to pimp it a little bit further:
- don't generate private/protected members
- exclude namespace detail with a doxygen annotation
...

So I would still give it a try.

jerryct:
As a starting point I added a Doxygen configuration:

https://github.com/jupp0r/prometheus-cpp/pull/189

You can run it with `doxygen doxygen.config` in the project source dir. I disabled many things in order to get only the public API. For example everything in namespace detail or directory detail is excluded. Can you try it with read the docs?

jerryct:
I think the easiest way to get the documentation is GitHub pages. We could simply create a new repository and let travis push to it:

https://github.com/jerryct/doc_test
https://jerryct.github.io/doc_test/index.html

PR #175, opened 2018-10-17
Document gauge builder

jerryct:


core/include/prometheus/gauge_builder.h
```diff
@@ -14,20 +14,61 @@ namespace detail {
...
+///
+/// The following metric and label conventions are not required for using
+/// Prometheus, but can serve as both a style-guide and a collection of best
+/// practices: https://prometheus.io/docs/practices/naming/
+///
+/// Finally the gauge family must be registered as a collector with
+/// GaugeBuilder::Register(Registry&). The gauge family fullfills the
+/// Collectable interface to expose the collected metrics. The registration is a
+/// key part needed to scrape metrics.
 detail::GaugeBuilder BuildGauge();
```
jerryct: During documenting the API I wondered why this function exists at all? Essentially the coding style

```
prometheus::BuildGauge()
                           .Name("some_name")
```
can also be realized with

```
detail::GaugeBuilder{}
   .Name(...)
   ...
```

It also brings me to the next point: Why is `GaugeBuilder` in namespace detail and thus an implementation detail? From the current usage it is an integral part of the API a user must use. So it should be either lifted into the prometheus namespace.

Or - if it is really desired to make `GaugeBuilder` an implementation detail - change `BuildGauge()` to something like `Family<Gauge>& RegisterGaugeFamily(labels, name, help, registry);` With the latter API the `GaugeBuilder` is truly an implementation detail because it is hidden from the user.

core/include/prometheus/gauge_builder.h
```diff
@@ -14,20 +14,61 @@ namespace detail {
...
 
 namespace detail {
 class GaugeBuilder {
  public:
+  /// \brief Configure a set of key-value pairs (= labels) of the metric.
```
jerryct: See my comment above. It is a little bit unfortunate to document a class in namespace detail for a user.

jerryct:
@gjasny What do you think about this?

https://github.com/jerryct/prometheus-cpp/commit/d84fafda4a65b5aa4469b06b6dc1d65a8f763418

We could leave this BuildCounter() function for backward compatibility, deprecate it or remove it as in this commit.

core/include/prometheus/gauge_builder.h
jerryct: But what is the benefit of this indirection vs. creating a Family and register the Family object?

But regardless of this, if the builders are an implementation detail I would move them to the include/detail directory:

https://github.com/jupp0r/prometheus-cpp/pull/214

jerryct:
@gjasny, @jupp0r, are you ok with this PR?

PR #177, opened 2018-10-18
Fix potential memory leak

jerryct:


core/src/registry.cc
```diff
@@ -2,40 +2,46 @@
...
+
 Family<Counter>& Registry::AddCounter(
     const std::string& name, const std::string& help,
     const std::map<std::string, std::string>& labels) {
   std::lock_guard<std::mutex> lock{mutex_};
-  auto counter_family = new Family<Counter>(name, help, labels);
-  collectables_.push_back(std::unique_ptr<Collectable>{counter_family});
-  return *counter_family;
+  collectables_.push_back(make_unique<Family<Counter>>(name, help, labels));
+  return dynamic_cast<Family<Counter>&>(*collectables_.back());
```
gjasny: Thanks for spotting this! I don't really like the dynamic_cast here. How about the following (untested):
```c++
auto counter_family = make_unique<Family<Counter>>(name, help, labels);
auto& ref = *counter_family;
collectables_.push_back(std::move(counter_family));
return ref;
```

jerryct: I think your approach also works although I struggled a little bit about storing a reference (line 2) from a object which is later moved-from.

I understand your concern about dynamic_cast  in general but I thought this is the most expressive form. The dynamic_cast can even be replaced by a static_cast because we know for sure that we pushed a Family<Counter>.  The dynamic_cast only checks this at runtime again. I thought I leave this extra check with dynamic_cast because these lines of code are not on the fastpath and maybe later someone refactors the code and forgets about this condition.

So if you still in favor of your snippet I will change it to your one.

gjasny: I'm wondering what will happen with `dynamic_cast` on the various platforms if RTTI is disabled. @jupp0r what's you opinion on the `dynamic_cast` vs. `auto& / move away`?

jerryct:
@gjasny  If you wish you can review and merge it.

gjasny:
Thank you for your dedication @jerryct !

PR #180, opened 2018-10-23
Allow people who have well defined CXXFLAG with -std=gnu++17 to use it safely if they want to

Romain-Geissler-1A:
Hi,

This fixes a mistake that unfortunately I see far too often in small cmake projects: -std=c++11 is being artibrarily forced to some values

Normally big organisations like mine do rebuild everything themselves and make sure that they control the C++ standard flags used all over all their compilation units. Indeed while quite close, C++11, C++14, C++17 and C++20 are both binary and source incompatible with each other. Compiling prometheus-cpp with C++11 while we actually use C++17 in all our own software otherwise may work, but is a good way to look for tricky problems to investigate in case things like STL containers passed from one C++17 binary to a C++11 binary are actually binary incompatible. This effect is even more true in presence of LTO builds (which we do as well) where it's very important that all compilation units have consistant flags with each other. So for these organisations who fully control the standard, let them use the good defaults already set in the environments CXXFLAGS.

Cheers,
Romain

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,14 +7,19 @@ include(GNUInstallDirs)
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 
-set(CMAKE_CXX_STANDARD 11)
-set(CMAKE_CXX_EXTENSIONS Off)
-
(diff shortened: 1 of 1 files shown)


gjasny:
Thanks for the report and patch!

PR #181, opened 2018-10-24
Add core and pull targets that don't link statically

palmerlao:


core/BUILD.bazel
```diff
@@ -11,3 +11,28 @@ cc_library(
...
     visibility = ["//visibility:public"],
 )
+
+cc_library(
+    name = "core-headers",
+    hdrs = glob(["include/**/*.h"]),
+    strip_include_prefix = "include",
+)
+
+cc_binary(
```
gjasny: Could you please tell me what's the reason behind separate `cc_binary` and `cc_library` targets?

pull/BUILD.bazel
```diff
@@ -15,3 +15,32 @@ cc_library(
         "@civetweb",
     ],
 )
```
gjasny: besides `core` and `pull` there's also a `push` library.

core/BUILD.bazel
palmerlao: My understanding is that the `cc_binary` target here only provides the shared object file. To use it in some other target, you'd put something like `//core:libcore.so` in the `srcs` field.

If you wrap `libcore.so` in a `cc_library`, you can both expose the headers and use something like `//core:core-shared` in the `deps` field instead. 

Personally, I find it a little awkward to have external targets in my `srcs` fields, but maybe that's not a popular opinion, or I could also have misunderstood how to use such targets.

pull/BUILD.bazel
palmerlao: I added a commit to try to do something similar, but I can't seem to build some of the `push` targets. I get the following error:

```
ERROR: /home/palmer/.cache/bazel/_bazel_palmer/6d05ea7d43fb77a9b934c1b9ffb77807/external/com_github_whoshuu_cpr/BUILD.bazel:4:1: C++ compilation of rule '@com_github_whoshuu_cpr//:cpr' failed (Exit 1)
external/com_github_whoshuu_cpr/cpr/error.cpp:3:23: fatal error: curl/curl.h: No such file or directory
```

Is there possibly a dependency on curl that hasn't been managed by Bazel?

gjasny:
If we'd remove `linkstatic = 1` from the libraries, then the targets decide if they'd like to use static or shared libs. That decision is made by `linkstatic = 1,` in the `cc_test` or `cc_binary` rule.

Would removing `linkstatic = 1` on the libs satisfy your needs?

pull/BUILD.bazel
gjasny: yes, libcurl is expected to be installed on the system.

palmerlao:
I gave that a try, along with using `linkstatic=0` in my `cc_binary` rule, but it seems that including the targets still triggers undefined reference errors during linking

gjasny:
Please re-open when necessary

PR #186, opened 2018-10-25
Move non public api

jerryct:


jerryct:
I don't change any code. I simply moved code to different files (my local diff tool gives a nicer diff).

jerryct:
I rebased it.

PR #188, opened 2018-10-25
Move benchmarks to top-level directory

jerryct:
I would like to propose to place the benchmarks in the top-level directory. Benchmarks are not really tests. This is also the way LLVM libcxx does it:

https://github.com/llvm-git-prototype/llvm/tree/master/libcxx
https://github.com/llvm-git-prototype/llvm/tree/master/libcxx/benchmarks

Or at least move it into the 'core' directory.


gjasny:
I fully agree that the benchmarks do not really belong to tests dir. But I'd prefer to keep them isolated in core, push, pull. It eases building and testing with bazel: You could simply say `bazel test //core/...` and avoid running bechmarks that depend on push or pull targets.

jerryct:
I moved the benchmarks to core/benchmarks.

PR #192, opened 2018-10-27
Add Doxygen config

jerryct:


jerryct:
Hi. I would like to keep the discussion open from https://github.com/jupp0r/prometheus-cpp/pull/167.

I think the easiest way to get the documentation is GitHub pages. We could simply create a new repository and let travis push to it:

https://github.com/jerryct/doc_test
https://jerryct.github.io/doc_test/annotated.html

The doxygen config is this:
https://github.com/jupp0r/prometheus-cpp/commit/2c476f175073617b5c3521af16c6589d17265d93

PR #194, opened 2018-10-27
Generate gh pages

gjasny:
@jupp0r  Could you please create a token named `GITHUB_TOKEN` as described [here](https://docs.travis-ci.com/user/deployment/pages/#setting-the-github-token) and add it to the global `secure` elements list in `.travis.yml`?

That should update doxygen on https://jupp0r.github.io/prometheus-cpp/
Example output available here: https://gjasny.github.io/prometheus-cpp/


.travis.yml
```diff
@@ -48,6 +48,18 @@ script:
...
+  - doxygen
+  - touch html/.nojekyll
+
+deploy:
+  provider: pages
+  local-dir: doc/html
+  skip-cleanup: true
+  github-token: $GITHUB_TOKEN
+  on:
+    branch: master
```
jerryct: I would not push to master and add generated files to the repository. The master branch is the one everybody checks out and works with. I would rather add another branch (like gh-pages) or simply create a second project under prometheus-cpp which only stores the generated files (like in my example).

.travis.yml
```diff
@@ -48,6 +48,18 @@ script:
   - make DESTDIR=`pwd`/deploy install
   - popd
 
+  - cd "${TRAVIS_BUILD_DIR}/doc"
```
jerryct: Better use `pushd` and call `popd` after line 53. That makes it extensible for the future without touching this 3 lines of code.

.travis.yml
```diff
@@ -48,6 +48,20 @@ script:
   - make DESTDIR=`pwd`/deploy install
   - popd
 
+  - pushd .
+  - cd doc
```
jerryct: only minor but you could write pushd ./doc and get rid of this extra cd

jupp0r:
Sorry for responding so late, got back from vacation. I'll take care of this later this week. Thanks @gjasny

jupp0r:
Should be unblocked by #250

jupp0r:
Please rebase + merge

PR #200, opened 2018-10-28
WIP Rename

jerryct:


jerryct:
This is work in progress but what do you think of the renaming in the last commit? I think this reflects the entities better.

PR #205, opened 2018-11-02
Clean third-party CMake files

jerryct:


cmake/civetweb-3rdparty-config.cmake
```diff
@@ -19,23 +19,18 @@ add_library(civetweb OBJECT
...
+    NDEBUG
+    NO_CGI
+    NO_CACHING
+    NO_SSL
+    NO_FILES
 )
 
-target_include_directories(civetweb PUBLIC ${CIVETWEB_INCLUDE_DIRS})
-
-if(BUILD_SHARED_LIBS)
```
jerryct: I removed because the target is an OBJECT target which does not build any library.

push/CMakeLists.txt
```diff
@@ -21,7 +18,6 @@ target_link_libraries(push
     ${PROJECT_NAME}::core
   PRIVATE
     ${CMAKE_THREAD_LIBS_INIT}
-    ${CURL_LIBRARIES}
```
jerryct: I moved all CURL references to the CPR package scripts, because the push library does not directly depend on it. Instead the CPR library has the dependency to CURL and should provide this as a transitive dependency.

cmake/FindCPR.cmake
```diff
@@ -1,3 +1,5 @@
+find_package(CURL REQUIRED)
```
gjasny: you should not set `REQUIRED` in a find module because it always makes the build fail if curl is not found. if cpr is not `required` but `optional` you do not want to fail. better remove the `required` here and add `CURL_FOUND` to FPHSA below.

cmake/civetweb-3rdparty-config.cmake
gjasny: you need to keep those `if(BUILD_SHARED_LIBS)` blocks because they affect how the obect library object files are compiled.

jerryct: Oh, I see. Sorry. I reverted it.

cmake/FindCPR.cmake
jerryct: Done.

PR #208, opened 2018-11-03
Squash Add() into one function

jerryct:


bdangit:
Whoa this is a pretty big change in the interface.  When you tag the release, I’d recommend a major bump in the version.

gjasny:
@bdangit I was thinking the same but it's in the `private` parts of Registry.

bdangit:
Ah! Yes. You are right. :thumbsup: In any case, would love to have a version bump.  :)

jupp0r:
it's in [v0.6.0](https://github.com/jupp0r/prometheus-cpp/releases/tag/v0.6.0)

PR #209, opened 2018-11-04
CMake improvements

gjasny:


core/CMakeLists.txt
```diff
@@ -20,7 +20,7 @@ add_library(${PROJECT_NAME}::core ALIAS core)
```
jerryct: Hi. You need to add `find_dependency(Threads)`. This is needed for the downstream projects because in `prometheus-cpp-targets.cmake` there is a line like this

```
set_target_properties(prometheus-cpp::pull PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_LIBRARIES "prometheus-cpp::core;\$<LINK_ONLY:Threads::Threads>;\$<LINK_ONLY:\$<\$<AND:\$<BOOL:UNIX>,\$<NOT:\$<BOOL:APPLE>>>:rt>>;\$<LINK_ONLY:\$<\$<BOOL:OFF>:ZLIB::ZLIB>>"
)
```

gjasny: Good catch!

PR #214, opened 2018-11-07
Move builder into detail directory

jerryct:


jerryct:
From the comment here https://github.com/jupp0r/prometheus-cpp/pull/175 shoudn't we place the builders into prometheus-cpp/detail? I put the BuildCounter() functions to the counter class but they could also moved to the family or registry. (registry is quiet good, because Register is the final function to call on BuildCounter()).

PR #216, opened 2018-11-10
Use Starlark http_archive and a widespread version of zlib

mapx:
This replaces the native http_archive and new_http_archive rules. The old rules will be disabled by default in a future version of Bazel (possibly 0.20).


gjasny:
Personally I find the GitHub URL be a more canonical name that zlib.net. This change will break opencensus build once again (cc @g-easy ).

Has someone tried to upstream a bazel build file to  zlib? I think it would also be pretty helpful to have a build file collection repo for non-bazelified upstream repos under the bazel umbrella project.

g-easy:
Thanks for CC'ing me, this did indeed break the OpenCensus build. Would you be willing to rename @net_zlib_zlib back to @com_github_madler_zlib to match gRPC?

The alternative seems to be an experimental bazel remapping feature: https://github.com/census-instrumentation/opencensus-cpp/pull/236

g-easy:
Rename proposal: https://github.com/jupp0r/prometheus-cpp/pull/225

PR #220, opened 2018-11-12
Travis: Use apt addon

gjasny:


jupp0r:
didn't know about this, nice!

PR #225, opened 2018-11-14
Rename net_zlib_zlib to com_github_madler_zlib to match gRPC.

g-easy:


gjasny:
For the record: That rename has been introduced with #216.

googlecartographer/cartographer seems to use `net_zlib_zlib` and census-instrumentation/opencensus-cpp likes `com_github_madler_zlib`. I'm sitting somehow between the chairs and have no strong preference on one of both choices.

Both projects seems to be Google driven. Maybe @mapx @g-easy @bogdandrutu @drigz @isturdy you could agree on a naming?

Is the `repo_mapping` working as expected?

mapx:
Nor do I have strong preference 😸. And I didn't mean to hurt anyone with that pull request😹.

drigz:
A couple of others: [rules_boost](https://github.com/nelhage/rules_boost/blob/2733bb5dddc0b62e794bb761a19e1bf67352bd7e/boost/boost.bzl#L86) uses net_zlib_zlib and [pubref/rules_protobuf](https://github.com/pubref/rules_protobuf/blob/99043441e3c473809f633f5ad049945606124b60/protobuf/build_file/com_github_madler_zlib.BUILD) uses com_github_madler_zlib.

I'd vote for net_zlib_zlib as it could remain "canonical" if the project stops using GitHub for hosting. But I don't have a strong preference, and agree that consistency would be best. If you accept this PR, I'm happy to send PRs to rules_boost and cartographer to change them.

g-easy:
@gjasny - opencensus-cpp doesn't directly depend on zlib, we depend on it transitively through gRPC. :)

@vjpai - what do you think (from gRPC's point of view) - how should we name zlib?

g-easy:
The new `http_archive` rule doesn't let me work around this with `repo_mapping`, I filed https://github.com/bazelbuild/bazel/issues/6845

Friendly ping: can we rename this back until there's a better solution?

vjpai:
Sorry for the late response to my mention (I think that was during the wildfires and I missed it). I don't see any reason why gRPC particularly cares what zlib is called, but I'll pass the mention to @nicolasnoble for all of our BUILD insights.

nicolasnoble:
The best way to solve this problem would be to get @madler to add a BUILD and WORKSPACE file. This would definitely stop any ambiguity. Anyone up to file a pull request there?

drigz:
@gjasny The path of least resistance seems to be to merge this and adjust cartographer. If anyone volunteers to add/maintain upstream Bazel support then at least we've saved them the trouble of picking a name :)

g-easy:
Good news is repo_mapping is back: https://github.com/bazelbuild/bazel/issues/6845 :)

drigz:
I've just noticed that some repositories refer to `//:z`, whereas others (tensorflow, rules_boost) refer to `//:zlib`. Cartographer uses an alias to provide both, presumably for compatibility with gRPC.

g-easy:
This is how opencensus-cpp closes the diamond: https://github.com/census-instrumentation/opencensus-cpp/blob/93a91fff4390cffdc4a640a61e1e93c0c3582097/WORKSPACE#L64

jupp0r:
Perfect ;)

jovany-wang:
Nice!
 Is this PR ready for merging?

g-easy:
Given https://github.com/jupp0r/prometheus-cpp/pull/225#issuecomment-463010996 I'm happy to withdraw this PR (or merge it. your (jupp0r) call.)

gjasny:
If if works for all parties, I'd leave it as-is.

PR #228, opened 2018-11-17
Avoid hash conflicts in some common cases.

jovan-wong:
For a hash function, we should avoid hash conflicts in some common cases, like the maps have the same combined string:
```c++
std::map<std::string, std::string> labels_1 {{"a", "bc"}};
std::map<std::string, std::string> labels_2 {{"ab", "c"}};
```
etc.

### DIFF
--- a/core/include/prometheus/family.h
+++ b/core/include/prometheus/family.h
@@ -187,7 +187,7 @@ std::size_t Family<T>::hash_labels(
         auto key_hash_code = std::hash<std::string>{}(label.first);
         final_hash_code = 31 * final_hash_code + key_hash_code;
 
-        auto value_hash_code = std::hash<std::string>{}(label.value);
+        auto value_hash_code = std::hash<std::string>{}(label.second);
         final_hash_code = 31 * final_hash_code + value_hash_code;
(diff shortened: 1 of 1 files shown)


gjasny:
Hi, thanks for pointing this issue out.

I'd like have some refactoring:
1) If it does not get too messy, the `hash_labels` function could be hidden from the user and implemented in a cpp file. It also does not depend on any template parameters so an standalone function would be possible (to not having to instantiate the necessary family templates).
2) An unit test for the expected family hashing would be nice
3) the hash combining function should live in its own header file
4) It should be written like `hash_combine` in [N3876](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n3876.pdf) But I don't have any idea how that document is licensed.

Cheers,
-Gregor

jovan-wong:
If I got what you said correctly, you‘d like me to do some refactoring about this, right?

gjasny:
Yes, that would be very helpful.

jovan-wong:
Okay.

gjasny:
Closing in favour of #235

PR #229, opened 2018-11-17
Use io forward declarations in header

jerryct:


push/src/gateway.cc
```diff
@@ -180,9 +182,7 @@ int Gateway::Delete() {
 }
 
 std::future<int> Gateway::AsyncDelete() {
-  return std::async(std::launch::async, [&] {
-    return Delete();
-  });
+  return std::async(std::launch::async, [&] { return Delete(); });
```
jovan-wong: Why change this line?
I think we should write an expression at independent line, right?

jerryct: My IDE automatically applies clang-format on the whole file. I added a new commit for the formatting correction.

jovan-wong: Could you please give me a url linked to the item in Google code style?

jerryct: I think this is it: https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace

But I cannot really say where it is written, but here is also lambda, which fits in one line: https://google.github.io/styleguide/cppguide.html#Formatting_Lambda_Expressions.

Essentially I don't really care because someone defined this style for clang-format :)
https://github.com/jupp0r/prometheus-cpp/blob/master/.clang-format

jovan-wong: It does make sense that clang-format could be defined by yourself.

Since there is no rule to require us write it inline by Google code style, we have no reason for changing it, right?

the current code of that line is also good right?

jovan-wong: Anyway, any way is okay.

jerryct: I think I don't understand your point. This project defines to adhere to the Google Style Guide: https://github.com/jupp0r/prometheus-cpp#contributing and provides the clang-format configuration file: https://github.com/jupp0r/prometheus-cpp/blob/master/.clang-format.

So there is no need to manually format anything because running `clang-format` on the changed files automatically applies the Google Style Guide.

jerryct:
@gjasny, @jupp0r, are you ok with this PR?

PR #230, opened 2018-11-19
Added support for shared object output

Spuddles:
I wanted to use the library in a project that already contained code built with the -fPIC option. Would it be possible to allow the building of process independent code to produce shared libraries?


gjasny:
Just add `-DCMAKE_BUILD_SHARED` to the command line. Does that work for you?

gjasny:
I'd like to avoid the two targets approach. It adds maintenance burden and seems unnecessary.

PR #236, opened 2018-12-06
Reformat everything

gjasny:
using clang-format version 8.0.0 (tags/google/stable/2018-10-04)


jovan-wong:
@gjasny 
How do you think of adding clang-format in CI to check our code style?
The CI will not pass if the committer  does not format his code style.

PR #237, opened 2018-12-07
Compile on build server with warnings as errors

jerryct:


pull/src/handler.cc
```diff
@@ -113,7 +113,7 @@ static std::size_t WriteResponse(struct mg_connection* conn,
   return body.size();
 }
 
-bool MetricsHandler::handleGet(CivetServer* server,
+bool MetricsHandler::handleGet(CivetServer* /*server*/,
```
gjasny: you could also write `std::ignore = server;` in the body. I always find the commented out parameters a liite bit clumsy.

core/include/prometheus/detail/utils.h
```diff
@@ -1,4 +1,4 @@
-#pragma onece
```
gjasny: good catch. don't know why I did not notice that.

pull/tests/integration/sample_server.cc
```diff
@@ -7,7 +7,7 @@
 #include <prometheus/exposer.h>
 #include <prometheus/registry.h>
 
-int main(int argc, char** argv) {
+int main(int /*argc*/, char** /*argv*/) {
```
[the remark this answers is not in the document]
jerryct: done

CMakeLists.txt
```diff
@@ -33,10 +33,17 @@ if(ENABLE_TESTING)
   enable_testing()
 endif()
 
-# suppress warnings
+# build flags
 
 add_compile_options(
   $<$<AND:$<STREQUAL:$<COMPILE_LANGUAGE>,CXX>,$<CXX_COMPILER_ID:AppleClang>>:-Wno-deprecated-declarations>
+  $<$<AND:$<STREQUAL:$<COMPILE_LANGUAGE>,CXX>,$<CXX_COMPILER_ID:AppleClang>,$<VERSION_LESS:$<CXX_COMPILER_VERSION>,9.3>>:-Werror>
```
gjasny: Could we guard that with a CMake option that will be enabled for CI? It is usually good practice to not use `-Werror` on the users compilers by default. Otherwise distributions will have to work-around it to not make their builds explode with newer compilers or uncommon platforms/architectures.

CMakeLists.txt
```diff
@@ -33,10 +33,17 @@ if(ENABLE_TESTING)
...
+  $<$<AND:$<STREQUAL:$<COMPILE_LANGUAGE>,CXX>,$<CXX_COMPILER_ID:AppleClang>,$<VERSION_LESS:$<CXX_COMPILER_VERSION>,9.3>>:-Wall>
+  $<$<AND:$<STREQUAL:$<COMPILE_LANGUAGE>,CXX>,$<CXX_COMPILER_ID:AppleClang>,$<VERSION_LESS:$<CXX_COMPILER_VERSION>,9.3>>:-Wextra>
```
gjasny: Is there a reason you limit the warnings to certain compiler versions?

pull/tests/integration/sample_server.cc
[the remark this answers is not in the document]
gjasny: Thanks!

pull/src/handler.cc
jerryct: I followed the way clang-tidy does it:
https://clang.llvm.org/extra/clang-tidy/checks/misc-unused-parameters.html

So if /* ... */ is too clumsy isn't it then better to omit the name altogether?

I now removed the unused parameter name but I can change it to std::ignore if you think this is better.

CMakeLists.txt
jerryct: New compiler versions normally add new warnings to -Wall -Wextra. My rationale is that the current Travis CI uses two specific compilers and especially GCC 4.8 is rather old. So we can only ensure for these two compilers that we don't have any warnings. If we would unconditionally switch on -Wall ... with -Werror a user of this library who might use a newer GCC or Clang version, may run into the problem that he cannot compile this library anymore.

Do you know of a better solution?

CMakeLists.txt
jerryct: yes, we could do it.

CMakeLists.txt
jerryct: If we guard the flags for the CI then the version number can be removed.

CMakeLists.txt
jerryct: done

CMakeLists.txt
jerryct: done. I removed the no-deprecation-declarations. Is it ok?

core/src/detail/hash.h
```diff
@@ -11,7 +11,7 @@ namespace detail {
 /// It's the boundary condition of this serial functions.
 ///
 /// \param seed Not effect.
-inline void hash_combine(std::size_t *seed) {}
```
jovan-wong: I think it's better to ignore `seed` in the body instead of remove `seed`.

core/src/detail/hash.h
```diff
@@ -1,4 +1,4 @@
-#pragma
```
jovan-wong: good catch!

core/tests/histogram_test.cc
```diff
@@ -11,7 +11,7 @@ TEST(HistogramTest, initialize_with_zero) {
   Histogram histogram{{}};
   auto metric = histogram.Collect();
   auto h = metric.histogram;
-  EXPECT_EQ(h.sample_count, 0);
```
jovan-wong: `EXPECT_EQ(h.sample_count, 0U);`  ---> EXPECT_EQ(0U, h.sample_count);
IIRC, the 1st argument of the function `EXPECT_EQ` means the expected value, and the 2nd one is real value.

core/tests/histogram_test.cc
```diff
@@ -21,7 +21,7 @@ TEST(HistogramTest, sample_count) {
   histogram.Observe(200);
   auto metric = histogram.Collect();
   auto h = metric.histogram;
```
jovan-wong: use `histogram` instead of the local variable `h`.

push/tests/integration/sample_client.cc
```diff
@@ -23,7 +23,7 @@ static std::string GetHostName() {
   return hostname;
 }
 
-int main(int argc, char** argv) {
```
jovan-wong: Why remove the 2 parameters?
I think it's unnecessary to do this change.

pull/src/handler.cc
```diff
@@ -113,8 +113,7 @@ static std::size_t WriteResponse(struct mg_connection* conn,
   return body.size();
 }
 
-bool MetricsHandler::handleGet(CivetServer* server,
-                               struct mg_connection* conn) {
```
jovan-wong: This change is unnecessary I think.

core/tests/family_test.cc
```diff
@@ -20,8 +20,8 @@ TEST(FamilyTest, labels) {
                          {{const_label.name, const_label.value}}};
   family.Add({{dynamic_label.name, dynamic_label.value}});
   auto collected = family.Collect();
-  ASSERT_GE(collected.size(), 1);
-  ASSERT_GE(collected.at(0).metric.size(), 1);
+  ASSERT_GE(collected.size(), 1U);
```
jovan-wong: I think the literal `1` is ok. It's unnecessary to use `1U`.

jerryct: I enabled warnings. size() returns an unsigned type and the warning complains about a compare between signed and unsigned.

core/tests/histogram_test.cc
jerryct: I only fix the compiler warnings and preserve the order as it was.

core/tests/histogram_test.cc
jerryct: This change is not related to the commit intention.

pull/src/handler.cc
jerryct: I enabled warnings. The warning complains about an unused parameter. server is not used in the body.

push/tests/integration/sample_client.cc
jerryct: argc and argv are not used in the function body.

jovan-wong: But for unused parameters, we should not remove them directly.
Just use `IGNORE` to ignore them.

pull/src/handler.cc
jovan-wong: Use `ignore` instead of removing it directly.

core/tests/family_test.cc
jovan-wong: Ok, I see.

core/include/prometheus/detail/utils.h
jovan-wong: @gjasny  Me too.

PR #240, opened 2018-12-10
configure number of threads used by Exposer

khvorov:
Exposer class creates two threads by default which could be too much in some cases. This pull-request allows to configure number of threads used by an exposer.


gjasny:
Thanks for your contribution

Issue #244, opened 2018-12-19
find civetweb not working as expected


gocarlos:
civetweb now exports the cmake files, so when you install civetweb we get `civetweb::civetweb` targets in the `usr/local/lib/cmake/civetweb`

those are different then the ones that we have here in the cmake folder: `CivetWeb::CivetWeb`

due to the lowercase vs camecase, I get an compilation (configuration) error when using the newest version of civetweb and the newest version of prometheus-cpp...

gocarlos:
cc/ @bachp

gjasny:
Unfortunately there's no release of civetweb, yet.

Could you please try to add the `MODULE` keyword to the civetweb `find_package()` command?

gjasny:
Due to the upcoming holiday season I'm short on time. Could you please file a pull request?

gocarlos:
Let me have a look on this tomorrow

gjasny:
Closed by #245

PR #245, opened 2018-12-20
fix(cmake): fix the target name of civetweb to be lowercase as upstream

gocarlos:
civetweb upstream now installs cmake files which can be used to do find_package here, though the targets upstream are all lower case... 

this PR fixes this mismatching

cc/ @bachp


gocarlos:
fixes: https://github.com/jupp0r/prometheus-cpp/issues/244

gocarlos:
@jupp0r @gjasny @bachp ready to review

gjasny:
Thanks for your help!

PR #247, opened 2018-12-21
Control symbol visibility and add SONAME

gjasny:


jupp0r:
Bazel is missing the exports header

gjasny:
I thought I could add windows dll support as well. But DLL, STL, and templates are a bad combination.

I probably need to hide some template implementations which might be a good thing for code isolation anyways.

PR #255, opened 2019-02-14
Cmake option for shared libs is BUILD_SHARED_LIBS with an "S" at the …

allavena:
Cmake option for shared libs is BUILD_SHARED_LIBS with an "S" at the end, as documented in issues/171


gjasny:
Thanks!

Issue #11, opened 2016-10-16
Add interface documentation


jupp0r:
All public interfaces need to be documented, especially regarding to memory management contracts and thread safety.

0mp:
Any suggestions on how/where we may want to document it?

jupp0r:
A good start would be in the headers.

gjasny:
We now have https://jupp0r.github.io/prometheus-cpp/ which is generated from the master branch.

PR #279, opened 2019-06-24
Add link to jupp0r.github.io/prometheus-cpp

mkenigs:
Was trying to find some docs and couldn't find anything until I found #11. Think it would be easier if there was a link in the readme


gjasny:
Thanks!

PR #285, opened 2019-07-03
core: Move elememts from one vector to the final one

gjasny:


jupp0r:
Nice optimization. I don't think the `end()` needs the `std::make_move_iterator`, but it doesn't hurt.

PR #286, opened 2019-07-06
core: Add option to merge families

gjasny:


core/include/prometheus/registry.h
```diff
@@ -52,19 +55,44 @@ class Registry : public Collectable {
...
+                          const std::map<std::string, std::string>& labels);
+  Family<Histogram>& AddHistogram(
+      const std::string& name, const std::string& help,
+      const std::map<std::string, std::string>& labels);
+  Family<Summary>& AddSummary(const std::string& name, const std::string& help,
+                              const std::map<std::string, std::string>& labels);
 
-  std::vector<std::unique_ptr<Collectable>> collectables_;
+  bool merge_families_;
+  std::vector<std::unique_ptr<Family<Counter>>> counters_;
```
gjasny: @jupp0r do you have a better idea how to store the not-type erased families w/o using std::variant and w/o using an ugly downcast?

gjasny: With C++17 it looks pretty OK: 71d9a18b6577131b3c69c2b173d2287db1a42ddf

core/include/prometheus/registry.h
```diff
@@ -36,6 +36,9 @@ namespace prometheus {
 /// a data race.
 class Registry : public Collectable {
  public:
+  explicit Registry(bool merge_families = false)
+      : merge_families_{merge_families} {}
```
jupp0r: I'd make that an enum

core/include/prometheus/registry.h
jupp0r: I don't think this is too bad, actually.

core/include/prometheus/family.h
```diff
@@ -90,6 +91,13 @@ class Family : public Collectable {
   Family(const std::string& name, const std::string& help,
          const std::map<std::string, std::string>& constant_labels);
 
+  bool IsSameAs(
```
jupp0r: We must only compare the name. There must only be one TYPE field per family:

> Only one TYPE line may exist for a given metric name.

That means that we need to either consolidate families with the same name, but different help messages/labels or error out.

[1] https://github.com/prometheus/docs/blob/master/content/docs/instrumenting/exposition_formats.md#text-format-details

gjasny: I added a check that the same family name does not show up with a different type. I'd like to work on the other should's and may's in a separate PR.

core/include/prometheus/registry.h
gjasny: done

core/src/registry.cc
```diff
@@ -1,17 +1,103 @@
...
+}  // namespace
+
+template <typename T>
+Family<T> &Registry::Add(const std::string &name, const std::string &help,
+                         const std::map<std::string, std::string> &labels,
+                         std::vector<std::unique_ptr<Family<T>>> &families) {
+  std::lock_guard<std::mutex> lock{mutex_};
+
+  if (insert_behavior_ == InsertBehavior::Merge) {
+    for (auto &family : families) {
```
jupp0r: use `std::find_if`

core/include/prometheus/registry.h
```diff
@@ -36,6 +36,20 @@ namespace prometheus {
 /// a data race.
 class Registry : public Collectable {
  public:
+  /// \brief How to deal with repeatedly added family names for a type
+  enum class InsertBehavior {
+    /// \brief Create new family object and append
```
jupp0r: Can we add another option, which should be the default: throw whenever trying to add a family with a name that already exists?

jupp0r: Can you describe more verbosely how the algorithm works? What happens when there is no existing family with that name, what happens when there is an existing family with that name, but a different type, etc?

jupp0r: ie in the doc comment for each behavior

gjasny: donw*e**

core/include/prometheus/family.h
jupp0r: ok

jupp0r:
Just realized that the commit message needs to be updated.

PR #290, opened 2019-08-05
Allow more than 1 histogram observation at a time

David-Ray1:
Added the ObserveMultiple method to the Histogram. This is useful
if you have already have values sorted into buckets, and the sum of
the values but no longer have the individual values.

### DIFF
--- a/core/include/prometheus/histogram.h
+++ b/core/include/prometheus/histogram.h
@@ -31,7 +31,7 @@ class Histogram {
 
   static const MetricType metric_type{MetricType::Histogram};
 
-  /// \brief Create a histogram with manually choosen buckets.
+  /// \brief Create a histogram with manually chosen buckets.
   ///
   /// The BucketBoundaries are a list of monotonically increasing values
   /// representing the bucket boundaries. Each consecutive pair of values is
@@ -50,6 +50,14 @@ class Histogram {
   /// b_n+1). The counter of the observed bucket is incremented. Also the total
   /// sum of all observations is incremented.
   void Observe(double value);
+
+  /// \brief Observe multiple data points.
+  ///
+  /// Increments counters given a count for each bucket. (i.e. the caller of
+  /// this function must have already sorted the values into buckets).
+  /// Also increments the total sum of all observations by the given value.
+  void ObserveMultiple(const std::vector<double> bucket_increments,
+                       const double sum_of_values);
 
   /// \brief Get the current value of the counter.
   ///
--- a/core/src/histogram.cc
+++ b/core/src/histogram.cc
@@ -24,6 +24,15 @@ void Histogram::Observe(const double value) {
   bucket_counts_[bucket_index].Increment();
 }
 
+void Histogram::ObserveMultiple(const std::vector<double> bucket_increments,
+                                const double sum_of_values) {
+  sum_.Increment(sum_of_values);
+
+  for (std::size_t i{0}; i < bucket_counts_.size(); ++i) {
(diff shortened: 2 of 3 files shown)


jupp0r:
Hey David, thanks for submitting this PR. Can you talk a little about the use case you have for this?

David-Ray1:
We have a product that processes packets, from which we want to output a Prometheus Histogram of the sizes of those packets. Calling the Observe method for each packet would be too slow. Instead we accumulate the packet size data using our own optimised mechanism. So we need to be able to insert that accumulated data into the histogram in one go.
I have also just noticed a bug with my ObserveMultiple method; you can give it a vector of the wrong length. I couldn't find a similar error handling example in Promethues-cpp, so how do you think it should be handled? Some suggestions:

- Check that the vector is of the right length, and if it's not throw an exception.

- Take any length of vector and if it's too short just increment the first n buckets, and if it's too long just ignore the extra elements.

- Make the method take the count for one bucket and a dummy value which it uses to work out which bucket to increment with the given count.

jupp0r:
> The performance requirements on the packet processing mean that calling Observe for each packet would cause too much of a performance hit

Have you measured this? I know of voip audio bridges that observe packet sizes for millions of packets per second. Conceptually, observing a value is finding the right bucket and incrementing a counter on the bucket, the number of values counter and the sum, which sounds like exactly what you do when you pre-aggregate those values. There's probably lots of potential of optimizing that code path in `Observe()`, too, especially the "find the right bucket" part.

David-Ray1:
Yes, we've observed the performance difference that calling it directly would incur.

When collecting the aggregated stats in our dataplane fast path, which handles hundreds of millions of packets per second, we collect the per packet data into a fixed number of fixed sized buckets, which are known in advance. This allows us to exploit custom, CPU-architecture-specific, heavily optimised code.

While I'm sure that enhancements to the performance of the Observe call path is possible, it's not possible to apply all of the same optimisations, while keeping the same flexible bucket count. In addition some of the function would require additional complexity in the library, and API, to avoid poor performance.

These optimisations feel like a poor fit for the library as is.

Do you have any specific objections to allowing the histogram to record multiple data values in parallel? In some applications, the individual data may not even be available?

jupp0r:
> Do you have any specific objections to allowing the histogram to record multiple data values in parallel? In some applications, the individual data may not even be available?

I just wanted to understand your use case really well to make sure that the effort we spend is spent in the right place. My first intuition was that `Observe()` should be fast enough for almost all use cases, but the points you brought up convinced me that it's indeed worthwhile to support your use case. Thanks for answering all my questions!

jupp0r:
> Thanks for answering all my questions!

And of course implementing this.

David-Ray1:
No problem.

I do still want to fix the bug I mentioned in my earlier comment, what solution do you think is best?

> Note, the ObserveMultiple method doesn't currently handle being provided a vector with the wrong number of elements. I think there are a number of approaches to resolve this, and it wasn't clear which error handling method best fit in with the library - there aren't any other similar examples.
> 
> Some suggestions:
> 
> * Check that the vector is of the right length, and if it's not throw an exception - e.g. a std::length_error - https://en.cppreference.com/w/cpp/error/length_error
> * Take any length of vector and if it's too short just increment the first n buckets, and if it's too long just ignore the extra elements.
> * Make the method take the count for one bucket and a dummy value which it uses to work out which bucket to increment with the given count.

jupp0r:
I'd opt for throwing. It's mostly a programmer error when this happens and the other two could lead to hard-to-find bugs.

PR #293, opened 2019-08-07
avoid infinite loop when counter is incremented from multiple threads

tom-code:
It seems that gauge/counter increment could cause infinite loop in case it is changed from different thread betwen value load and compare_exchange_weak operation.
When compare_exchange_weak fails new value must to be read to be able to succeed at some point.


gjasny:
From `std::atomic<T>::compare_exchange_weak( T& expected, T desired , ...)`:

> Atomically compares the object representation (until C++20) value representation (since C++20) of *this with that of expected, and if those are bitwise-equal, replaces the former with desired (performs read-modify-write operation). Otherwise, loads the actual value stored in *this into expected (performs load operation).

So `expected` is updated if modification fails

PR #294, opened 2019-08-08
Templatize builders

mkenigs:
Was looking at the builders and didn't see any reason they aren't templatized, unless I missed something. Tried to change it and looks like the only real difference it would have is on includes, since registry doesn't have to include all the metrics anymore. Let me know how it looks @jupp0r


jupp0r:
LGTM, definitely removes some duplication.

PR #296, opened 2019-08-14
Fix Histogram::ObserveMultiple bug. Now throw length_error if the num…

David-Ray1:
…ber of bucket increments is not equal to the number of buckets in the histogram.


jupp0r:
Thanks for following up on this David.

PR #322, opened 2019-12-19
CP-1664: modify prometheus-cpp

fcimeson:
The prometheus-cpp lib provides functionality to expose internal data to Prometheus via a series of sockets (unique to each node). Typically, each metric will get exposed for the lifetime of the process (node). Thus, if a node created 1000 metrics at the beginning of the program and never updated them again for the entire process, prometheus-cpp would continue exposing them to the socket for the entire process (similar to a while loop with 1000 print lines).

Additionally, prometheus-cpp uses a number of return by references, which poses the following issue:
```
auto& family = BuildCounter()....
auto& metric = family.Add(labels);
family.Remove(metric);
metric.Increment(); // Segfault

```
This can be fixed by returning a shared pointer instead. However, the nuance to this is that a shared pointer that has been removed may update its data but that data will not get exposed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ int main(int argc, char** argv) {
                              .Name("time_running_seconds_total")
                              .Help("How many seconds is this server running?")
                              .Labels({{"label", "value"}})
+                             .RetentionBehavior(prometheus::RetentionBehavior::Keep)
                              .Register(*registry);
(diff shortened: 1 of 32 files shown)


fcimeson:
PR to merge to wrong branch... (sorry)

PR #323, opened 2019-12-19
Make Collect methods const

BenjaminW3:
Collect methods should not influence the state of the objects in any
externally visible way.

### DIFF
--- a/core/include/prometheus/collectable.h
+++ b/core/include/prometheus/collectable.h
@@ -19,7 +19,7 @@ class PROMETHEUS_CPP_CORE_EXPORT Collectable {
   virtual ~Collectable() = default;
 
   /// \brief Returns a list of metrics and their samples.
-  virtual std::vector<MetricFamily> Collect() = 0;
+  virtual std::vector<MetricFamily> Collect() const = 0;
 };
 
 }  // namespace prometheus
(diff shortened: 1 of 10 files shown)


core/src/family.cc
```diff
@@ -69,7 +69,7 @@ const std::map<std::string, std::string> Family<T>::GetConstantLabels() const {
 }
 
 template <typename T>
-std::vector<MetricFamily> Family<T>::Collect() {
+std::vector<MetricFamily> Family<T>::Collect() cont {
```
jupp0r: typo here

BenjaminW3: fixed

core/include/prometheus/collectable.h
```diff
@@ -19,7 +19,7 @@ class PROMETHEUS_CPP_CORE_EXPORT Collectable {
   virtual ~Collectable() = default;
 
   /// \brief Returns a list of metrics and their samples.
-  virtual std::vector<MetricFamily> Collect() = 0;
+  virtual std::vector<MetricFamily> Collect() const = 0;
```
gjasny: Changing the exposed ABI demands a version bump in root CMakLists.txt to 0.9.0

gjasny:
Build could be fixed by appending a `@v1` to `uses: actions/checkout`

BenjaminW3:
Thanks for fixing the internal mutability issues!
The coverage build still fails.

PR #331, opened 2020-01-29
Avoid copying the histogram metric buckets.

dgavedissian:
Thanks for the really nice to use library!

Whilst running `heaptrack` against a project that depends on it, I noticed we were allocating/free'ing a lot of memory somewhere. I tracked this down to a loop which was calling `ObserveMultiple`, passing a persistent `std::vector` object each time. I realised that the vector was being copied rather than being passed by reference.

Looking at the implementation of `ObserveMultiple`, it appears that passing by reference was the intention, and it was just a typo to miss the `&`, so this PR adds the missing ref qualifiers.


jupp0r:
Good catch and thanks for the PR. I'll fix CI and will merge it ASAP and release a new version.

PR #365, opened 2020-05-12
Add support for serving multiple paths

jameseh96:
In large projects it may be desired to have sets of stats with different scrape intervals and retention periods - currently that would require two `Exposer`s on different ports (AFAICT?).

This change shouldn't break existing usage of `Exposer` (but probably changes the ABI), but provides the option of having multiple endpoints with potentially different groups of `Collectable`s. These can then be served on a single civet server, on a single port.

If this is out of scope no worries, or if needs a version number bump let me know.

### DIFF
--- a/pull/CMakeLists.txt
+++ b/pull/CMakeLists.txt
@@ -14,6 +14,7 @@ if(ENABLE_COMPRESSION)
 endif()
 
 add_library(pull
+  src/endpoint.cc
   src/exposer.cc
   src/handler.cc
   src/handler.h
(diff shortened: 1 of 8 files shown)


pull/CMakeLists.txt
```diff
@@ -14,6 +14,7 @@ if(ENABLE_COMPRESSION)
 endif()
```
gjasny, line 14: Could you please rebase to latest master to get bazel builds running?

gjasny:
I'll have a deeper look within the next days

jameseh96:
Just spotted #369 - happy to close this. Thanks @gjasny

gjasny:
Hope it will fit your needs. I wanted to take @jupp0r a look before merging

PR #372, opened 2020-06-04
Add support for HTTP Basic Auth

jameseh96:
Prometheus supports Basic Auth for [scrape targets](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config).

This PR also adds a 3rd party base64 dependency. I've tried to match the existing thirdparty structure when adding it, but happy to change it if it doesn't match the preferred style.


gjasny:
Thanks for you PR. I took a look at the build errors and started working on them. Hopefully I'll finish the work during the next days.

Did you know that certificate-based authentication was already possible?

PR #390, opened 2020-09-10
Avoid race when registering a collectable

mlsna:
A thread registering a `Collectable` with the `Exposer` may end up
invalidating iterators to the `collectables_` vector inside `Endpoint`
while a `CivetServer` handler thread is iterating the vector.

Move the `collectables_` vector to `MetricsHandler` where all accesses
can be guarded by a mutex.

### DIFF
--- a/pull/src/endpoint.h
+++ b/pull/src/endpoint.h
@@ -30,7 +30,6 @@ class Endpoint {
  private:
   CivetServer& server_;
   const std::string uri_;
-  std::vector<std::weak_ptr<Collectable>> collectables_;
   // registry for "meta" metrics about the endpoint itself
   std::shared_ptr<Registry> endpoint_registry_;
   std::unique_ptr<MetricsHandler> metrics_handler_;
--- a/pull/src/endpoint.cc
+++ b/pull/src/endpoint.cc
@@ -11,8 +11,8 @@ Endpoint::Endpoint(CivetServer& server, std::string uri)
     : server_(server),
       uri_(std::move(uri)),
       endpoint_registry_(std::make_shared<Registry>()),
-      metrics_handler_(detail::make_unique<MetricsHandler>(
-          collectables_, *endpoint_registry_)) {
+      metrics_handler_(
+          detail::make_unique<MetricsHandler>(*endpoint_registry_)) {
   RegisterCollectable(endpoint_registry_);
   server_.addHandler(uri_, metrics_handler_.get());
 }
@@ -24,7 +24,7 @@ Endpoint::~Endpoint() {
 
 void Endpoint::RegisterCollectable(
     const std::weak_ptr<Collectable>& collectable) {
-  collectables_.push_back(collectable);
+  metrics_handler_->RegisterCollectable(collectable);
 }
 
 void Endpoint::RegisterAuth(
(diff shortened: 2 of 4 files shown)


mlsna:
The coverage check failed with:
```
Thu, 10 Sep 2020 23:49:08 GMT Total Test time (real) =  18.61 sec
Thu, 10 Sep 2020 23:49:09 GMT Traceback (most recent call last):
Thu, 10 Sep 2020 23:49:09 GMT   File "/home/runner/.local/bin/coveralls", line 11, in <module>
Thu, 10 Sep 2020 23:49:09 GMT     sys.exit(run())
Thu, 10 Sep 2020 23:49:09 GMT   File "/home/runner/.local/lib/python2.7/site-packages/cpp_coveralls/__init__.py", line 92, in run
Thu, 10 Sep 2020 23:49:09 GMT     raise ValueError("\nno coveralls.io token specified and no travis job id found\n"
Thu, 10 Sep 2020 23:49:09 GMT ValueError: 
Thu, 10 Sep 2020 23:49:09 GMT no coveralls.io token specified and no travis job id found
Thu, 10 Sep 2020 23:49:09 GMT see --help for examples on how to specify a token
Thu, 10 Sep 2020 23:49:09 GMT
Thu, 10 Sep 2020 23:49:09 GMT ##[error]Process completed with exit code 1.
```

From looking at other PRs, this seems to be a recurring issue. Please let me know if there is any action I should take with regard to the coverage check failure.

jupp0r:
> Please let me know if there is any action I should take with regard to the coverage check failure.

No action needed, thanks for finding and fixing this. I guess for most use cases we create and register all Collectables before creating the Exposer so this never occurred.

jupp0r (approved):
LGTM, maybe @gjasny could have a second look.

jupp0r:
> From looking at other PRs, this seems to be a recurring issue.

Thanks for pointing this out, the issue is that github does not share repository secrets with PRs originating from forks (which makes sense, if you modified the scripts you could steal the secrets). I created #391 to not run test coverage uploads in these situations, which should fix the situation. Again thanks for pointing this out.

pull/src/endpoint.h
```diff
@@ -30,7 +30,6 @@ class Endpoint {
  private:
   CivetServer& server_;
   const std::string uri_;
-  std::vector<std::weak_ptr<Collectable>> collectables_;
```
gjasny, line 33: Wasn't the whole point of supporting multiple endpoints that those could serve different collectables?

gjasny (changes requested):

pull/src/endpoint.h
mlsna: As I understand, this PR preserves that functionality.

A single `Exposer` contains one or more `Endpoint` objects, depending on the `uri` parameter(s) used when calling `Exposer::RegisterCollectable`. And each of those `Endpoint` objects contains its own `MetricsHandler`, responsible for scraping the corresponding `Endpoint::collectables_` vector.

This change preserves the cardinality of those relationships, just moving the `collectables_` vector down a level, from `Endpoint` to `MetricsHandler`, where the locking can be done more in a more straightforward matter.

mlsna: @gjasny: Is there still any concern regarding the handling of multiple endpoints?

PR #391, opened 2020-09-11
Fix coverage from forks

jupp0r:
Understandably, pull requests originating in forks won't get access to secrets in github (in order to prevent them being extracted). This change makes coverage upload only run from PRs or pushes originating in this repo, getting rid of failures for external contributors.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -20,7 +20,7 @@ jobs:
       - name: Prepare
         run: .github/scripts/run-prepare ${{ matrix.buildsystem }} ${{ matrix.os }}
       - name: Test
-        if: ${{ github.repository == "jupp0r/prometheus-cpp" }}
+        if: github.repository == 'jupp0r/prometheus-cpp'
         env:
           COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/coverage.yml
```diff
@@ -20,6 +20,7 @@ jobs:
       - name: Prepare
         run: .github/scripts/run-prepare ${{ matrix.buildsystem }} ${{ matrix.os }}
       - name: Test
+        if: github.repository == "jupp0r/prometheus-cpp"
```
gjasny, line 23: There's a parsing error in the green (?!?) coverage check. Maybe the condition has to be surrounded by `${{ }}`
```suggestion
        if: ${{ github.repository == "jupp0r/prometheus-cpp" }}
```

.github/workflows/coverage.yml
```diff
@@ -20,6 +20,7 @@ jobs:
...
         env:
           COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
```
gjasny, line 25: Maybe it would be cleaner to move this env block up to the job level and check for the presence of the `COVERALLS_REPO_TOKEN` variable in the environment.

jupp0r: Yeah I thought so too but if conditions don't work on jobs :(

PR #405, opened 2020-10-21
Generate shlib information in debian package.

davidlenfesty:
Produces a more correct debian package. This is almost essential if you're building a debian package with prometheus-cpp as a dependancy. (see [CPACK_DEBIAN_PACKAGE_GENERATE_SHLIBS](https://cmake.org/cmake/help/v3.6/module/CPackDeb.html#variable:CPACK_DEBIAN_PACKAGE_GENERATE_SHLIBS))


gjasny:
Thank you!

PR #408, opened 2020-11-06
Fix insufficiently precise value rendering

### LINKED ISSUE #407: Precision loss in float rendering
We happened to be looking at this code as part of OpenMetrics, and noticed that there appears to be precision loss in how floating points are rendered by this client library:

https://github.com/jupp0r/prometheus-cpp/blob/3a8ed550e9fa805394e4506de5a1cd205163ef6f/core/src/text_serializer.cc#L13

The default precision for C++ floats is 6, which is not sufficient for a 64bit floating point which can have 17 significant digits. I believe this be changed to explicitly ask for 17 digits of precision.

jupp0r:
This change fixes #407 by increasing precision for rendered floating point numbers
output in the text serializer.

### DIFF
--- a/core/src/text_serializer.cc
+++ b/core/src/text_serializer.cc
@@ -21,7 +21,7 @@ void WriteValue(std::ostream& out, double value) {
     oldState.copyfmt(out);
 
     out.setf(std::ios::fixed, std::ios::floatfield);
-    out << std::setprecision(17);
+    out << std::setprecision(std::numeric_limits<double>::max_digits10);
     out << value;
 
     out.copyfmt(oldState);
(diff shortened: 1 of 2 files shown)


core/src/text_serializer.cc
```diff
@@ -16,9 +17,14 @@ void WriteValue(std::ostream& out, double value) {
   } else if (std::isinf(value)) {
     out << (value < 0 ? "-Inf" : "+Inf");
   } else {
-    auto saved_flags = out.setf(std::ios::fixed, std::ios::floatfield);
+    std::ios oldState{nullptr};
+    oldState.copyfmt(out);
+
+    out.setf(std::ios::fixed, std::ios::floatfield);
+    out << std::setprecision(17);
```
gjasny, line 24: Maybe you could use `std::numeric_limits<double>::max_digits10` instead of the magic number:

https://stackoverflow.com/questions/554063/how-do-i-print-a-double-value-with-full-precision-using-cout

PR #417, opened 2020-11-23
pull: Use internal base64 decoding function to avoid dependency

gjasny:


pull/tests/internal/BUILD.bazel
```diff
@@ -0,0 +1,14 @@
+cc_test(
+    name = "internal",
+    srcs = glob([
+        "*.cc",
+        "*.h",
+    ]),
+    copts = ["-Iexternal/googletest/include"],
+    linkstatic = True,
+    deps = [
+        "//pull",
```
gjasny, line 10: link against core not pull

jupp0r (approved):

Issue #421, opened 2020-12-10
Adding summary-family or gauge-family with wrong chars in name forbid adding other ones


FabienCarmagnac:
v0.10.0

Code to reproduce , very easy :

```
Exposer exposer{"127.0.0.1:8080"};
auto registry = std::make_shared<Registry>();
exposer.RegisterCollectable(registry);

// pipes ok, now add stuff
auto& counter0 = BuildCounter().Name("this_one_will_work").Register(*registry).Add({{ "el", "re"}});
auto& counter1 = BuildCounter().Name("fa mi ly with space in name or |").Register(*registry).Add({{ "toto", "tata"}});
auto& counter2 = BuildCounter().Name("foo").Register(*registry).Add({{ "toto", "tata"}});
auto& counter3 = BuildCounter().Name("bar").Register(*registry).Add({{ "toto", "tata"}});

```

==> family foo and bar will never show in prometheus server because family with name="fa mi ly with space in name or |" (counter1) broke something (http parsing ?).

PR #426, opened 2020-12-13
core: enforce proper metric and label names

gjasny:
Closes: #421


jupp0r:
LGTM, exceptions are great here because this is mostly programmer error when labels are wrong.

Issue #441, opened 2020-12-30
The quantile name Not very good for Summary/Histogram metrics


bigc2000:
test_duration_ms{instance="localhost:8080", job="run", module="run", quantile="0.75000000000000000"}
test_duration_ms{instance="localhost:8080", job="run", module="run", quantile="0.94999999999999996"}
test_duration_ms{instance="localhost:8080", job="run", module="run", quantile="0.98999999999999999"}

but compate to offical metrics:
prometheus_tsdb_wal_fsync_duration_seconds{instance="localhost:9090", job="prometheus", quantile="0.5"}
prometheus_tsdb_wal_fsync_duration_seconds{instance="localhost:9090", job="prometheus", quantile="0.9"}
prometheus_tsdb_wal_fsync_duration_seconds{instance="localhost:9090", job="prometheus", quantile="0.99"}

The quantile lable is too long for double value,  and if LabelNames provided may be much better.

gjasny:
What do you mean with too long? Could you please be more verbose and describe what you expect and what is broken right now?

bigc2000:
> What do you mean with too long? Could you please be more verbose and describe what you expect and what is broken right now?
 **quantile**="0.75000000000000000"   expected "0.75" ,other than "0.75000000...".

I modified the code, may be n

ot very good. 
 
- Add label for summay's quantile 

-Set precision(2)  for quantile's default lable 

A diff details:
`diff --git a/core/tests/summary_test.cc b/core/tests/summary_test.cc
index f902fbb..9f5ed0d 100644
--- a/core/tests/summary_test.cc
+++ b/core/tests/summary_test.cc
@@ -44,12 +44,13 @@ TEST(SummaryTest, quantile_size) {
 
 TEST(SummaryTest, quantile_bounds) {
   Summary summary{Summary::Quantiles{
-      {0.5, 0.05, "0.5"}, {0.90, 0.01, "0.90"}, {0.99, 0.001, "0.99"}}};
+      {0.50, 0.05, "0.50"}, {0.90, 0.01, "0.90"}, {0.99, 0.001, "0.99"}}};
+  
   auto metric = summary.Collect();
   auto s = metric.summary;
   ASSERT_EQ(s.quantile.size(), 3U);
-  EXPECT_EQ(s.quantile.at(0).quantile, "0.5");
-  EXPECT_EQ(s.quantile.at(1).quantile, "0.9");
[...]

PR #444, opened 2021-01-10
fix: use printf %g or std::to_chars to serialize doubles

gjasny:
This will result in floating point numbers redered with the
necessary precision and avoids trailing zeroes.

Issue: #441


draveness:
I think this could be very useful, can we merge this @jupp0r :)

qinix:
+1 for this

jupp0r (approved):

PR #470, opened 2021-03-20
push: force use of POST in case if body is empty

### LINKED ISSUE #469: If body is empty when pushing 'GET' is used and result is 405 instead of 200
If all the metrics are being created dynamically there is a moment when there are no metrics registered.
If you push at this moment body will be empty and GET method will be used instead of POST/PUT.
Using GET method results in 405 response instead of 200 (if you use empty body with POST).

My suggestion is to use `curl_easy_setopt(curl, CURLOPT_POST, 1);`

aldoshkind:
fix #469

### DIFF
--- a/push/src/gateway.cc
+++ b/push/src/gateway.cc
@@ -92,6 +92,7 @@ int Gateway::performHttpRequest(HttpMethod method, const std::string& uri,
 
   curl_easy_reset(curl);
   curl_easy_setopt(curl, CURLOPT_URL, uri.c_str());
+  curl_easy_setopt(curl, CURLOPT_POST, 1);
 
   curl_slist* header_chunk = nullptr;
 


gjasny:
But doen't this change semantics? See: [pushgateway#put-method](https://github.com/prometheus/pushgateway#put-method)
> POST works exactly like the PUT method but only metrics with the same name as the newly pushed metrics are replaced (among those with the same grouping key).

aldoshkind:
I am not shure I understand exactly what do you mean by "change semantics".
Semantics of `Gateway::Push` is changed and I consider it as an improvement: now this method does push if no metrics provided instead of GET request, which is not defined for push operations in document you mentioned. Thus, behavior of `Gateway::Push` is more conformant to spec now.

Suggested change does only one thing - it forces POST to be the default method, so behavior of `performHttpRequest` in case of PUT and DELETE is not affected. If `performHttpRequest` is called with `method` = PUT or DELETE, corresponding option wil be set explicitly few line below (I double checked :) )

gjasny:
Obsolete with #471

Issue #178, opened 2018-10-23
dynamic creation and deletion of counters.


sunildr:
I have a use-case where my component (object) which creates the prometheus counters is dynamically created when a connection comes up. And when the connection goes down, the component is deleted. 

What happens to the counters and the registry when the holding object is deleted? Is there a way to unregister/deregister the counters from the registry before my component is deleted?

The registry is not part of this component i.e. it stays even without the component.

jupp0r:
Does that make sense?

hhoffstaette:
So this affects me too, even if it's with low cardinality. In my case I'm trying to write an exporter for filesystems, and even though those have unique fsids (effectively limiting the cardinality to at most a dozen) they can be mounted/unmounted, and I need to stop collecting & updating metrics for filesystems that have disappeared. I understand the point you made in #183 (use a map to manage my own sets of counters) but I still need a way to unregister Collectables from the registry again.
The only other way would be to stop the scraper & recreate everything - which seems unnecessarily complicated, not to mention fragile. A simple Delete() that takes a collection (for batching) of the refs returned from Add() would be super useful.

jupp0r:
Just to give you a heads up, I'm implementing this right now and should have something ready for review in the next few days.

laozhang007:
This affects me too.Hope to delete all counters or registry at once and free the memory.

brian-brazil:
> In my case I'm trying to write an exporter for filesystems, and even though those have unique fsids (effectively limiting the cardinality to at most a dozen) they can be mounted/unmounted, and I need to stop collecting & updating metrics for filesystems that have disappeared

This doesn't sound right, you wouldn't be deleting metrics in that case but removing children. It sounds like what you're looking for is a custom collector, which appears to be the `Collectable` interface in this client library.

hhoffstaette:
> This doesn't sound right, you wouldn't be deleting metrics in that case but removing children.
> It sounds like what you're looking for is a custom collector, which appears to be the `Collectable`
> interface in this client library.

You'll be glad to hear that's exactly what I did back when. :)

gjasny:
The `Exposer` nowadays has a `RemoveCollectable` function:
https://github.com/jupp0r/prometheus-cpp/blob/77a5b0172833786ec014498d61a4dd87c500dc3b/pull/include/prometheus/exposer.h#L34-L35

CoderPoet:
> The `Exposer` nowadays has a `RemoveCollectable` function:
[...]

PR #482, opened 2021-04-21
add method to query family if metric already exists

wose:
If metrics are dynamically added and removed to and from a family (i.e. in a RAII way), there must only be only owner of the reference returned by the `family.Add()` method. To allow this we need to know if a metric with specific labels already exist within the family or likely end up with dangling references.

This relates to #178

### DIFF
--- a/core/include/prometheus/family.h
+++ b/core/include/prometheus/family.h
@@ -119,6 +119,11 @@ class PROMETHEUS_CPP_CORE_EXPORT Family : public Collectable {
   /// if the given metric was not returned by Add().
   void Remove(T* metric);
 
+  /// \brief Returns true if the dimensional data with the given labels exist
+  ///
+  /// \param labels A set of key-value pairs (= labels) of the dimensional data.
+  bool Has(const std::map<std::string, std::string>& labels) const;
+
   /// \brief Returns the name for this family.
   ///
   /// \return The family name.
--- a/core/src/family.cc
+++ b/core/src/family.cc
@@ -77,6 +77,13 @@ void Family<T>::Remove(T* metric) {
 }
 
 template <typename T>
+bool Family<T>::Has(const std::map<std::string, std::string>& labels) const {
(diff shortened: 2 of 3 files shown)


jupp0r:
The ownership model is pretty simple: there is only one owner of any one metric: the registry. Non-owning references to metrics in the registry are returned at creation time and there can be as many references as you want. The only one responsible (and actually allowed) to free the memory taken up by the Metric is the Registry.

In light of that, I don't understand the rationale outlined in the commit message. Your change looks ok in the sense that it is correct, but can you elaborate a little bit more on your use case?

wose:
I'll try to better outline my use case. Let's say I have a few metrics as part of another object, those metrics should be added when that object is being constructed and removed when it's destroyed. I have a wrapper similar to the following to deal with this:

``` cpp
template<typename T>
class prom_wrapper final {
public:
    template<typename ...Args>
    prom_wrapper(prometheus::Family<T> & family, std::map<std::string, std::string> const & labels, Args && ...args)
        : family(family)
        , metric(&family.Add(labels, std::forward<Args>(args)...)) { }

    prom_wrapper(prom_wrapper && o)
        : family(o.family)
        , metric(std::exchange(o.metric, nullptr)) { }

    ~prom_wrapper() {
        if (metric) {
            family.Remove(metric);
        }
    }

    T * operator->() { return metric; }
    // some parts omitted
private:
    prom_wrapper(prom_wrapper const &) = delete;
    prom_wrapper & operator=(prom_wrapper const &) = delete;

    prometheus::Family<T> & family;
    T * metric;
};
```

To make this work I need to make sure there will never be more than one `prom_wrapper` with the same family and labels. Else the first one destroyed will leave the others with a dangling reference. With the proposed change I can enforce that during runtime by checking this requirement during construction of the `prom_wrapper` and throw.

jupp0r (approved):

jupp0r:
Thanks for the description of your use case.

PR #483, opened 2021-05-01
fix make_unique

pavel-pimenov:


pull/src/handler.cc
```diff
@@ -151,7 +151,7 @@ bool MetricsHandler::handleGet(CivetServer*, struct mg_connection* conn) {
     metrics = CollectMetrics(collectables_);
   }
 
-  auto serializer = std::unique_ptr<Serializer>{new TextSerializer()};
+  auto serializer = detail::make_unique<TextSerializer>();
```
gjasny, line 154: There's no need for the Text Serializer to be created on the heap. Please create an instance on the stack.

gjasny (approved):

gjasny (changes requested):
Please remove
```
../pull/src/handler.cc should remove these lines:
- #include "prometheus/serializer.h"  // lines 19-19
```
as reported by include-what-you-use.

Afterwards we can merge your change!

Thanks,
Gregor

gjasny (approved):

PR #486, opened 2021-05-07
CMakeLists: Change cmake min version to 3.14

amaier17:
For cmake versions older than 3.14, the cpack generation fails at the
top level directory with the following:

CMake Error: The source directory "" does not exist.
Specify --help for usage, or press the help button on the CMake GUI.

Using cmake 3.14 or newer fixes this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 
-cmake_minimum_required(VERSION 3.12 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.14 FATAL_ERROR)
 if(POLICY CMP0091)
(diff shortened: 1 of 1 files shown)


gjasny:
Hello,

Looking at the stable Ubuntu and Debian (backports) repo, we should be able to even go up to 3.16. But 3.14 is fine as well.

Thanks for your contribution!
-Gregor

gjasny (approved):

amaier17:
Thanks Gregor. I found this while compiling on a Centos8 deployment which I think by default is 3.11 (for cmake3). Probably best to keep the minimum version down at 3.14 until there are features of >3.16 that you're using here as compiling the newest versions can be somewhat tedious on older deployments.

Thanks for the merge!

Issue #211, opened 2018-11-05
prometheus zlib versus grpc zlib


g-easy:
https://github.com/jupp0r/prometheus-cpp/commit/6bc9b933d7eab3104a89919871dcaa217b440255 adds `@com_github_madler_zlib//:zlib` whereas gRPC already has `@com_github_madler_zlib//:z`

This causes a collision on repository name, but different target names, and breaks the build for https://github.com/census-instrumentation/opencensus-cpp (which pulls in both prometheus and gRPC)

PR #507, opened 2021-08-05
rename library z to zlib

hiberabyss:
Rename zlib label name to compatable with other rules like `https://github.com/nelhage/rules_boost`


gjasny:
Hello,

Unfortunately, it's not that easy. See: #211.

Thanks,
Gregor

gjasny:
This is another thread talking about zlib and Bazel:
https://github.com/jupp0r/prometheus-cpp/pull/225#issuecomment-438815835

gjasny:
Hopefully that issue could be resolved once and for all with bzlmod.

PR #520, opened 2021-09-11
Support to hide metrics which hasn't update for a long time

acelyc111:


acelyc111:
Hi @gjasny , could you please take a look at this PR? thanks

fluffy:
Hi @acelyc111 - I was just reading your code in this PR and it looks like exactly what I need. Is this code working well for you. I'm considering just pulling your PR into my fork. Thanks.

acelyc111:
> Hi @acelyc111 - I was just reading your code in this PR and it looks like exactly what I need. Is this code working well for you. I'm considering just pulling your PR into my fork. Thanks.

@fluffy Yes it works well, feel free to try it.

gjasny:
Hello,

I see some problems with this pull request:
1) Setting `last_update_ts_` is not synchronized between the updating thread and the collecting thread. A 32 bit system setting a 64 bit time_t might read funny values.

2) std::time is not monotonic and might step backwards.

3) The metric types have been designed for low-overhead when setting/updating. Looking at the clock every time adds a huge overhead. The included benchmark shows some numbers (on my M1 Mac Mini)

```
Before:
BM_Counter_Increment                              10.3 ns         10.3 ns     55964630

After:
BM_Counter_Increment                              45.3 ns         45.3 ns     15390579
```

Maybe a smarter algorithm could shave-off most of the overhead while also providing thread safety.

@jupp0r Do you have any comments?

Thanks,
Gregor

core/src/family.cc
```diff
@@ -108,7 +109,10 @@ std::vector<MetricFamily> Family<T>::Collect() const {
   family.type = T::metric_type;
   family.metric.reserve(metrics_.size());
   for (const auto& m : metrics_) {
-    family.metric.push_back(std::move(CollectMetric(m.first, m.second.get())));
+    if (!m.second->Expired(current_ts)) {
+      family.metric.push_back(
+          std::move(CollectMetric(m.first, m.second.get())));
```
sjanel, line 114: This `std::move` is useless here

PR #528, opened 2021-10-14
fix: there is a race condition when pushing and collecting histogram data

t-wynants:
The rest of the library is written in a thread-safe way so I guess the non-thread-safe code in the histogram is a bug.

All increment counters and the total sum can be changed by another threads while the collection thread is preparing the cumulative data collection.  
This will result in an invalid scrape that can even contain pink elephants.


gjasny:
Thanks for you contribution!

PR #536, opened 2021-11-07
Rule of Five applied on Gateway, avoid unnecessary copies of std::string

sjanel:
Pass `std::string` by reference instead of copy (is there a reason why we copy strings here that I missed?).
Explicitly delete special members of `Gateway` class to respect rule of five


gjasny:
Thanks for you contribution!

Issue #555, opened 2022-01-14
gatway.Push will block my program


DeserterW:
if I didn't create  metrics for collector, my program will stopped right there. I wonder know if there is any possible do not how to create gateway successfully without metrics, because sometimes, I need to create gateway first so I can add metrics later.

anatoly-p:
got the same issue… curl_wrapper handles empty data on POST improperly… need to set CURLOPT_POSTFIELDSIZE to 0, otherwise curl is waiting for chunked data to be provided, but wrapper doesn't do that

PR #556, opened 2022-02-07
fix(push): Allow POST empty data

anatoly-p:
Fixed: #555


gjasny:
Thank you fix debugging and fixing it.

gjasny:
Do you need a new release with the changes?

anatoly-p:
no urgency… for now I can use the library without the fix just POSTing only non-empty data… that's OK if it will be released with other fixes in upcoming release

PR #563, opened 2022-03-28
Add new constructor for Histogram and Summary taking rvalue ref of parameters

sjanel:
- Set default values to object fields in header file to limit code duplication as well.
- Mark constructors taking only one parameter as explicit

### DIFF
--- a/core/include/prometheus/histogram.h
+++ b/core/include/prometheus/histogram.h
@@ -44,7 +44,9 @@ class PROMETHEUS_CPP_CORE_EXPORT Histogram {
   /// exponential etc..
   ///
   /// The bucket boundaries cannot be changed once the histogram is created.
-  Histogram(const BucketBoundaries& buckets);
+  explicit Histogram(const BucketBoundaries& buckets);
+
+  explicit Histogram(BucketBoundaries&& buckets);
 
   /// \brief Observe the given amount.
   ///
@@ -68,7 +70,7 @@ class PROMETHEUS_CPP_CORE_EXPORT Histogram {
   ClientMetric Collect() const;
 
  private:
-  const BucketBoundaries bucket_boundaries_;
+  BucketBoundaries bucket_boundaries_;
   mutable std::mutex mutex_;
   std::vector<Counter> bucket_counts_;
   Gauge sum_;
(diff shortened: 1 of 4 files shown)


core/src/summary.cc
```diff
@@ -20,13 +23,13 @@ void Summary::Observe(const double value) {
 }
 
 ClientMetric Summary::Collect() const {
-  auto metric = ClientMetric{};
```
gjasny, line 23: Whats wrong with those? We tru to follow the [almost-always-auto](https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/) style here.

sjanel: Nothing wrong, it's just simpler and not involving `auto` nor assignment. I have not heard about this style, thanks for the link. In the end, I guess it's just a matter of style, personally I am not a fan of this but no problem to revert it to keep style consistent.

core/include/prometheus/histogram.h
```diff
@@ -68,10 +70,10 @@ class PROMETHEUS_CPP_CORE_EXPORT Histogram {
   ClientMetric Collect() const;
 
  private:
-  const BucketBoundaries bucket_boundaries_;
+  BucketBoundaries bucket_boundaries_;
   mutable std::mutex mutex_;
   std::vector<Counter> bucket_counts_;
-  Gauge sum_;
+  Gauge sum_{};
```
gjasny, line 76: I think you could omit the `{}` here. The `sum_` member was only present in the member initializer list to show the order of members.

core/src/summary.cc
```diff
@@ -4,14 +4,17 @@
...
+                 int age_buckets)
     : quantiles_{quantiles},
-      count_{0},
-      sum_{0},
       quantile_values_{quantiles_, max_age, age_buckets} {}
 
-void Summary::Observe(const double value) {
+Summary::Summary(Quantiles&& quantiles, std::chrono::milliseconds max_age,
+                 int age_buckets)
+    : quantiles_(std::move(quantiles)),
```
gjasny, line 14: Could you make that curly braces as well or will it get ambiguous?

gjasny (changes requested):

core/src/summary.cc
gjasny: Thanks. I'd like to keep those as-is.

gjasny (approved):

PR #564, opened 2022-03-28
Binary search for bucket boundaries in Histogram::Observe

sjanel:


gjasny:
Before:
```
--------------------------------------------------------------------
Benchmark                          Time             CPU   Iterations
--------------------------------------------------------------------
BM_Histogram_Observe/0          53.1 ns         53.1 ns     13167264
BM_Histogram_Observe/1          53.1 ns         53.1 ns     13160368
BM_Histogram_Observe/8          54.1 ns         54.1 ns     12907746
BM_Histogram_Observe/64         56.8 ns         56.8 ns     12329153
BM_Histogram_Observe/512         140 ns          140 ns      4993508
BM_Histogram_Observe/4096        704 ns          704 ns       994869
```

After:
```
--------------------------------------------------------------------
Benchmark                          Time             CPU   Iterations
--------------------------------------------------------------------
BM_Histogram_Observe/0          53.1 ns         53.1 ns     13168801
BM_Histogram_Observe/1          53.1 ns         53.1 ns     13132469
BM_Histogram_Observe/8          53.1 ns         53.1 ns     13177959
BM_Histogram_Observe/64         53.0 ns         53.0 ns     13180192
BM_Histogram_Observe/512        53.4 ns         53.4 ns     13089495
BM_Histogram_Observe/4096       61.0 ns         61.0 ns     11408083
```

core/src/histogram.cc
```diff
@@ -17,13 +17,11 @@ Histogram::Histogram(const BucketBoundaries& buckets)
...
 }
 
-void Histogram::Observe(const double value) {
-  // TODO: determine bucket list size at which binary search would be faster
-  const auto bucket_index = static_cast<std::size_t>(std::distance(
-      bucket_boundaries_.begin(),
-      std::find_if(
-          std::begin(bucket_boundaries_), std::end(bucket_boundaries_),
-          [value](const double boundary) { return boundary >= value; })));
+void Histogram::Observe(double value) {
```
gjasny, line 20:
```suggestion
void Histogram::Observe(const  double value) {
```

Could you please restore the `const`? The author uses it to show that the passed value is not modified within the function. In case we decide to remove it it should be done everywhere.

gjasny:
Please rebase. CI scripts here still use the `windows-2016` agent which is not available, anymore.

core/src/histogram.cc
sjanel: Sure, done

sjanel:
> Please rebase. CI scripts here still use the `windows-2016` agent which is not available, anymore.

done

PR #569, opened 2022-03-30
Set clang-format to always left align references and pointers

ljbade:
While working on our fork I noticed an annoying issue where clang-format wasn't consistently left aligning references and pointers. I only noticed this as our own code uses right aligned style so I was doing that by habit until I noticed the a mix of alignment.

It seems by default `clang-format` will try to derive the alignment based on what it finds already in a file. I changed the `.clang-format` to force always left aligned and it found a few files with mixed alignment.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,1 +1,2 @@
 BasedOnStyle: Google
+DerivePointerAlignment: false
--- a/core/src/detail/hash.h
+++ b/core/src/detail/hash.h
@@ -11,14 +11,14 @@ namespace detail {
 /// It's the boundary condition of this serial functions.
 ///
 /// \param seed Not effect.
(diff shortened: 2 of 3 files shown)


.clang-format
```diff
@@ -1 +1,3 @@
 BasedOnStyle: Google
+DerivePointerAlignment: false
+PointerAlignment: Left
```
gjasny, line 3: Could you please remove the explicit `PointerAlignment` here? It seems to be the default:
```
$ clang-format --style=Google -dump-config|grep PointerAlign
DerivePointerAlignment: true
PointerAlignment: Left
```

ljbade: ok

gjasny:
Thank you for the contribution!

PR #570, opened 2022-03-30
Add missing std:: on size_t

ljbade:
Noticed two files where `size_t` was used without the `std::` namespace.


gjasny:
Thank you for your contribution!

PR #571, opened 2022-03-30
Fix IWYU issues

ljbade:
I was checking some changes in our fork with IWYU and it seems to pick up a lot of issues with master.

I want to check that I am using the correct version of IWYU in case if it an issue of old version. The version I am using is the one in Ubuntu 20.04:
```
iwyu --version
include-what-you-use 0.12 based on clang version 9.0.1-10 
```

Otherwise this PR fixes most of the issues it finds except for two that I can't figure out how to fix. I am not very familiar with this tool.

The remaining issues:
```
/home/leith/swift/prometheus-cpp/pull/include/prometheus/exposer.h should add these lines:
#include "/home/leith/swift/prometheus-cpp/pull/src/endpoint.h"  // for Endp...

/home/leith/swift/prometheus-cpp/push/include/prometheus/gateway.h should add these lines:
#include "/home/leith/swift/prometheus-cpp/push/src/curl_wrapper.h"
```

### DIFF
--- a/core/include/prometheus/detail/builder.h
+++ b/core/include/prometheus/detail/builder.h
@@ -6,6 +6,7 @@
 
 // IWYU pragma: private
 // IWYU pragma: no_include "prometheus/family.h"
+// IWYU pragma: no_include "prometheus/registry.h"
 
 namespace prometheus {
 
--- a/core/include/prometheus/registry.h
+++ b/core/include/prometheus/registry.h
@@ -11,12 +11,16 @@
 #include "prometheus/labels.h"
 #include "prometheus/metric_family.h"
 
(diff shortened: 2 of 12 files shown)


gjasny:
Hello,

thank for the PR. Unfortunately, the IWYU 0.12 is very buggy. Therefore we're using 8.17 (from Debian Sid) in CI. But also there sometimes it tries to flap things back-and forth.

Thanks,
Gregor

ljbade:
Ah OK, I will not worry about this check for the moment then. Thanks for letting my know I need a very recent version!

gjasny:
Once Ubuntu 22.04 is released upgrading to a working version will be easier.

PR #579, opened 2022-04-06
fix(core): Ignore labels with empty value

gjasny:
The Data Mpdel spec states:
> A label with an empty label value is considered
> equivalent to a label that does not exist.


BenjaminW3:
Conceptually this looks good to me.

I am not sure about the performance impact. In our codebase we have some places where we dynamically do `family.Add({{"key", value}}).Increment();` on each observation. This would now end up with an additional copy of the labels.

PR #586, opened 2022-05-18
Add third mirror for zlib

krfricke:
The second mirror is not working anymore. At the moment, `mirror.bazel.build`'s SSL certificate is expired. This adds another mirror to make zlib available again.

cc @gjasny @jupp0r


gjasny:
Thank you!

PR #595, opened 2022-06-08
feat(core): Add Info metric

petitlapin:
Hi,

Info is a metric type defined in the [OpenMetrics format](https://github.com/OpenObservability/OpenMetrics/blob/98ae26c87b1c3bcf937909a880b32c8be643cc9b/specification/OpenMetrics.md#info).

It allows to use labels for static information that will not change during the lifetime (for example, versions of the executable/librairies, compilation date...).

In prometheus, it is emulated using a [Gauge with a value 1](https://github.com/prometheus/prometheus/blob/2ce94ac19673a3f7faf164e9e078a79d4d52b767/cmd/promtool/testdata/metrics-test.prom#L33-L35).

After checking, it is not implemented in all the clients, but I've found it in the [Java](https://github.com/prometheus/client_java/blob/2b60c479f57a2f1e20a5bab4dc7131df42327da7/simpleclient/src/main/java/io/prometheus/client/Info.java), [Python](https://github.com/prometheus/client_python#info) and [Rust](https://github.com/prometheus/client_rust/blob/master/src/metrics.rs#L20) implementations.

Cheers,

Johnny

ps: I couldn't run iwyu yet, I have issues installing it.

### DIFF
--- a/core/src/text_serializer.cc
+++ b/core/src/text_serializer.cc
@@ -96,6 +96,13 @@ void SerializeGauge(std::ostream& out, const MetricFamily& family,
   WriteTail(out, metric);
 }
 
+void SerializeInfo(std::ostream& out, const MetricFamily& family,
+                   const ClientMetric& metric) {
+  WriteHead(out, family, metric, "_info");
+  WriteValue(out, metric.info.value);
+  WriteTail(out, metric);
+}
+
 void SerializeSummary(std::ostream& out, const MetricFamily& family,
                       const ClientMetric& metric) {
   auto& sum = metric.summary;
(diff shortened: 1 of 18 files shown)


core/src/text_serializer.cc
```diff
@@ -164,6 +171,14 @@ void SerializeFamily(std::ostream& out, const MetricFamily& family) {
         SerializeGauge(out, family, metric);
       }
       break;
+    // info is not handled by prometheus, we use gauge as workaround
+    // (https://github.com/OpenObservability/OpenMetrics/blob/98ae26c87b1c3bcf937909a880b32c8be643cc9b/specification/OpenMetrics.md#info-1)
+    case MetricType::Info:
+      out << "# TYPE " << family.name << " gauge\n";
```
petitlapin, line 177: For reference, the [Java]((https://github.com/prometheus/client_java/blob/70e0f5ab5bf8f8876a18c75d33d1c9dc4cac9232/simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java#L188-L189)) and [Python](https://github.com/prometheus/client_python/blob/c835d17749f53f19387aba09b419e4cc28eca320/prometheus_client/exposition.py#L204-L206) implementations. However, [Rust](https://github.com/prometheus/client_rust/blob/eef56cf5792694e6ca03c0193c59675a4d3622e1/src/encoding/text.rs#L749) seems to use info as type.

gjasny:
Thanks, your change looks good. Hopefully I’ll find some time next week to look at it again and merge it.

PR #601, opened 2022-07-12
Add Reset APIs to histogram and counters

hkadayam:
These reset apis are specifically useful when histograms are collected from different way and then convert them to prometheus metric.


hkadayam:
Any comments about this PR? Thanks

gjasny:
Thank you! And sorry that it took a while.

PR #602, opened 2022-07-24
chore(core): Converted raw for-loop to `std::find_if`

tcoyvwac:
Hi there & Greetings :smile_cat:!
Wanted to help project out by converting a raw for-loop to its standard algorithm equivalent.


dmitrmax:
It seems that std::remove_if instead of find_if+erase would be better

gjasny:
Hello,
due to some customers who still are on C++11 we cannot use the `auto` keyword in the lambda. And without that the lambda looks rather clumsy. Converting to `remove_if` and `erase_if` has the downside that the loop is not aborted on first hit.

I'd keep the code as-is.

Thanks,
Gregor

Issue #618, opened 2022-11-28
Can not build with c++11 only(gcc 4.8)


owent:
Can not build with c++11 only(gcc 4.8)
OS: Linux

gjasny:
The master requires C++14 (see #615). Please rely on the 1.0 branches.

owent:
I create a PR #619  to fix this problem, maybe it's helpful and I have already test it under ubuntu 18.04 and gcc 4.8.

gjasny:
Why can't you use C++14?

owent:
> Out of curiosity: why can't you use C++14?

GCC 4.8 does not implement all libraries of C++14.

Issue #615, opened 2022-11-13
Add C++11 support with a make_unique workaround


juandemanjon:
I need to use prometheus-cpp in a build system that only supports C++11 and utilizing std::make_unique forces me to create a patch on top of this wonderful library.

It will be easier if the library itself recognizes that it is building in C++11 and uses it owns make_unique function like 
```cpp
/**
 *  std::make_unique replacement for C++11
 *
 *  @param args the arguments to make_unique (forwarded)
 *
 *  @return the unique pointer
 */
template <typename T, typename... Args>
std::unique_ptr<T> make_unique(Args&&... args) {
  return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
}
}
```
source: https://gist.github.com/chinmaygarde/970fd5bbd124754b7d36

gjasny:
Please use the 1.0.x releases for that. I'll try to backport any bugfixes to them.

PR #620, opened 2022-11-28
build: revert back to C++11

gjasny:
There seem to be many poor souls stuck to C++11. Because we don't really depend on C++14 features let's wait for a reason to bump up the standard to C++17 or later.

Close: #619
Fix: #618
Fix: #615


juandemanjon:
#615 Thank you!

PR #622, opened 2022-12-08
Use std::atomic::fetch_add in modern C++.

Romain-Geissler-1A:
Small contribution, hoping that it might be a bit faster when using recent compilers/C++ dialiects.


Romain-Geissler-1A:
I have just rebased my pull request ;)

PR #626, opened 2022-12-22
Update platform constraints to be compatible with bazel 6

vihangm:
Bazel 6 flipped the incompatible_use_platforms_repo_for_constraints flag
which causes `@bazel_tools//platforms:*` constraints to fail. The fix is to
use the constraints from `@platforms//*`.

See https://github.com/bazelbuild/bazel/issues/8622 for more details.

Signed-off-by: Vihang Mehta <vihang@pixielabs.ai>


gjasny:
Thanks for your contribution.

PR #640, opened 2023-03-06
build(cmake): rpm package via custom spec file

synarete:
Provide a detailed RPM spec file and use it to generate rpm package via cmake build system. Although CPack knows how to create rpm packages, it produces low quality spec file. Instead, use dedicated spec file and generate rpm packages with custom target:

  $ cmake -B_build -DENABLE_RPM_SPEC=ON -DBUILD_SHARED_LIBS=ON
  $ cmake --build _build --target rpm

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,6 +26,7 @@ option(THIRDPARTY_CIVETWEB_WITH_SSL "Enable SSL support for embedded civetweb so
 option(OVERRIDE_CXX_STANDARD_FLAGS "Force building with -std=c++11 even if the CXXFLAGS are configured differently" ON)
 option(GENERATE_PKGCONFIG "Generate and install pkg-config files" ${UNIX})
 option(RUN_IWYU "Run include-what-you-use" OFF)
+option(ENABLE_RPM_SPEC "Create rpm package using custom spec" OFF)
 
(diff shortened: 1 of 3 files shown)


gjasny:
Hello,

could you please elaborate on the "low quality" of the CPack generated rpm package? What are you missing?

Thanks,
Gregor

synarete:
> Hello,
> 
> could you please elaborate on the "low quality" of the CPack generated rpm package? What are you missing?
> 
> Thanks, Gregor

When generating rpm package via CPack's rpm generator, we get a spec file which has poor or missing fileds:
1) License set to `unknown`
2) Group set to `unknown`
3) %description is minimal
4) No %build section
5) %files section uses hard-coded path instead of macro-based
6) Does not have `Requires` and `BuildRequires`

In addition, it creates a single big package instead of separated packages (such as `prometheus-cpp`, `prometheus-cpp-devel` and `prometheus-cpp-debuginfo` as proposed in this PR). Why would one want to have prometheus-cpp's header-files installed on target machine, where all that is needed is the shared-library?

gjasny:
I'm sorry but creating a perfect RPM package is out of scope for this library. I'd suggest you either try to get prometheus-cpp into Fedora or create a [COPR](https://copr.fedorainfracloud.org/) or [OBS](https://build.opensuse.org/) recipe.

PR #651, opened 2023-05-15
Encode empty push label values with '=' character.

### LINKED ISSUE #652: Push does not work with blank grouping key values
Pushgateway encodes grouping label keys and values as URL path segments. Those values may be blank, but URL path segments may not be blank. To get around this, pushgateway requires that empty values be encoded as the character '='.

See https://github.com/prometheus/pushgateway#url for details.

This project doesn't do this and leaves the value blank, so if you have a label with a name like "label1" and a blank value, you get a request path like this: `/metrics/job/label1//label2/bla`. That's not a valid request path, or worse, it is, but it's equivalent to `/metrics/job/label1/label2/bla`, which doesn't work.

BrandonY:
Pushgateway encodes grouping label keys and values as URL path segments. Those values may be blank, but URL path segments may not be blank. To get around this, pushgateway requires that empty values be encoded as the character '='. See https://github.com/prometheus/pushgateway#url for details.

Fixes #652

### DIFF
--- a/push/src/gateway.cc
+++ b/push/src/gateway.cc
@@ -49,7 +49,10 @@ void Gateway::RegisterCollectable(const std::weak_ptr<Collectable>& collectable,
 
   if (labels) {
     for (auto& label : *labels) {
-      ss << "/" << label.first << "/" << label.second;
+      // A single = is required to encode an empty label value into a URL,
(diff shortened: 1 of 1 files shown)


gjasny:
Closing in favour of #675.

PR #653, opened 2023-06-15
feat(push): allow timeout for HTTP requests

OwenKnight-FLINTpro:
Allows specification of a timeout for HTTP requests which is passed to curl. This allows the operation to eventually fail if the remote endpoint does not respond.

Default value for the added parameter maintains existing behaviour.


gjasny:
Merged as #676

PR #656, opened 2023-08-16
INF-4014: add pkgconfig file generation

nightingale0131:
Updated CMakeLists to include pkconfig args for cmake.
Updated make_deb.sh to generate deb with depedency info.


nightingale0131:
Oops, didn't mean to merge into upstream..

PR #683, opened 2023-12-28
Avoid temporary objects

gjasny:


gjasny:
@owent, @esigo, @lalitb: FYI: this PR would change the `Collectable` signature.

Did you ever run into problems with excessive number of metrics? It's nice to see that prometheus-cpp can cope with 2 million counters. But do such setups exist in the wild?

owent:
> @owent, @esigo, @lalitb: FYI: this PR would change the `Collectable` signature.
> 
> Did you ever run into problems with excessive number of metrics? It's nice to see that prometheus-cpp can cope with 2 million counters. But do such setups exist in the wild?

Is there any way to tell otel-cpp whether to use the new class `Serializer` to get `Serializer`? I found no macro to check the ABI changes.

PR #706, opened 2024-03-20
Add presetupCurl callback to Gateway.

nbelov-magdv:
I use self-signed certificate, so I need possibility to add extra options for curl.

For example, for disabling check ssl certificate I do:

```	
const prometheus::Labels Labels = prometheus::Gateway::GetInstanceLabel("https://pushgateway.prometheus.localhost");
prometheus::Gateway Gateway{
    "https://pushgateway.prometheus.localhost", "", "sample_client", Labels, {}, {}, {},
    [&](CURL* curl)
    {
        curl_easy_setopt(curl, CURLOPT_SSL_VERIFYPEER, 0);
    }
};
```

### DIFF
--- a/push/include/prometheus/gateway.h
+++ b/push/include/prometheus/gateway.h
@@ -28,8 +28,11 @@ class PROMETHEUS_CPP_PUSH_EXPORT Gateway {
   Gateway(const std::string& host, const std::string& port,
           const std::string& jobname, const Labels& labels = {},
           const std::string& username = {}, const std::string& password = {},
-          std::chrono::seconds timeout = {},
-          std::function<void(CURL*)> presetupCurl = nullptr);
+          std::chrono::seconds timeout = {});
+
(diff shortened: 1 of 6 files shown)


push/include/prometheus/gateway.h
```diff
@@ -24,7 +28,8 @@ class PROMETHEUS_CPP_PUSH_EXPORT Gateway {
   Gateway(const std::string& host, const std::string& port,
           const std::string& jobname, const Labels& labels = {},
           const std::string& username = {}, const std::string& password = {},
-          std::chrono::seconds timeout = {});
```
gjasny, line 27: Up to now we provided an abstraction to curl. But I can see the necessity to set arbitrary options.

Could you please add a second ctor with your presetup hook that does not have the `username`, `password`, and `timeout` parameters? Those could all be set within the new hook.

PR #707, opened 2024-03-28
Added two cmake cross compile toolchain file.

shiptux:
This commit includes RISC-V and AArch64 cross-compile support for Prometheus-cpp.

Full steps:

```bash
git submodule init
git submodule update

mkdir aarch64_build
cd aarch64_build

# run cmake with cross-compile toolchain
cmake .. -DBUILD_SHARED_LIBS=ON -DENABLE_PUSH=OFF -DENABLE_COMPRESSION=OFF -DCMAKE_TOOLCHAIN_FILE=../cmake/toolchain-aarch64.cmake

# build
cmake --build . --parallel 4

# test
Test should runs on QEMU. WIP

### DIFF
--- a/cmake/toolchain-aarch64.cmake
+++ b/cmake/toolchain-aarch64.cmake
@@ -1,0 +1,71 @@
+# **********************************************************
+# Copyright (c) 2014-2021 Google, Inc.    All rights reserved.
+# **********************************************************
+
+# Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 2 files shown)


gjasny:
Hello,

prometheus-cpp is not an application but a library. Because all C++ components should be built with the same toolchain, I don't believe that prometheus-cpp is the proper place to store those.

Thanks,
Gregor

PR #709, opened 2024-05-04
Use curl as bazel_dep

mmorel-35:
Curl has been publish as a bzlmod.
This use it as a bazel_dep instead of bazel extension

I also used buildifier to format MODULE.bazel and placed dev_dependencies after build dependencies


gjasny:
Thank you!

PR #713, opened 2024-05-14
Add config for basic package version file

PythonGermany:
Added CMake config to automatically generate version information file. This makes it possible to check if the correct version of the package is installed when the package is consumed.

I'm not sure if the COMPATIBILITY parameter is the one that fits this library best. 
If you like this change and want to merge it just change it to the value you want.


gjasny:
Thank you for the contribution. I triggered a CI run.

PR #731, opened 2025-02-24
remove locks in some degree for the histogram

FAKERINHEART:
remove locks in Histogram::Observe and Histogram::ObserveMultiple


FAKERINHEART:
@jupp0r @gjasny hi, will this pull request be taken for cr?

gjasny:
Sorry, most of the tests fail right now. In the current form I won't be able to maintain the code because I have no idea what it is doing, what the design idea behind it is and what the limitations are. E.g.: will it still work on weakly ordered architectures?

PR #742, opened 2025-08-09
Allow injection of external CivetServer via Exposer constructor

### LINKED ISSUE #716: Non-collectable resources in Exposer
Hi!

We would like to serve additional non-Prometheus resources under the same HTTP server (e.g. typical health check (```/healthz```) and readiness (````/readyz```) endpoints) along with the metrics resources, without loosing the current functionality provided by the ```Exposer```class.

One possible solution would be using ```std::shared_ptr``` instead of ```std::unique_ptr``` to store the ```server_``` member and having an additional constructor that accepts the provision of an externally managed CivetServer.

Thank you in advance.

aeroyorch:
Resolves #716

### DIFF
--- a/core/include/prometheus/detail/future_std.h
+++ b/core/include/prometheus/detail/future_std.h
@@ -6,11 +6,14 @@
 namespace prometheus {
 namespace detail {
 
-// Remove as soon C++14 can be used.
+#if __cplusplus >= 201402L
+using std::make_unique;
+#else
 template <typename T, typename... Args>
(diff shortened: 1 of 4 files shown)


aeroyorch:
Hello @gjasny ,
I’ve submitted this PR, which implements issue #716. When you have a chance, I’d really appreciate your review.  
Thanks in advance for your support!

aeroyorch:
Hi @gjasny , if you could take a look on this very appreciated. Thanks!

jupp0r (approved):
LGTM

Issue #636, opened 2023-02-20
docstrings are outdated/incorrect


samin36:
The docstring for the ``Add`` method in ``family.h`` mentions a ``std::runtime_exception`` exception is thrown if invalid label values are passed in, but this is incorrect. Looking at ``family.cc`` reveals the type thrown is ``std::invalid_argument``.

FujiwaraNaoto:
I'm new to this community and I'd like to fix this issue.

PR #744, opened 2025-09-20
fix incorrect document for  family.h

FujiwaraNaoto:
I fixed incorrect description, which is mentioned [here](https://github.com/jupp0r/prometheus-cpp/issues/636).

`Family()` constructor  and `Add()` method raises `std::invalid_argument` exception, but the document `family.h` says `\throw std::runtime_exception on invalid metric or label names.`


FujiwaraNaoto:
Could you review it?

gjasny:
Thanks!

PR #747, opened 2026-04-30
Build the project on Bazel 9.x.

pgold:
A few packages were upgraded, and the reliance on built-in C++ rules was dropped.


gjasny:
Thanks! Do you need a new release with those changes?

pgold:
Thanks for the prompt review!

I don't *need* one (I apply these patches locally), but I think releasing a new version would help (Bazel build) users migrating to Bazel 9.x.

PR #750, opened 2026-06-04
Use `std::less<>` for `Labels` to enable heterogeneous lookup on C++14+

aeroyorch:


gjasny:
Hello,

do you have a benchmark that shows improvements with this change? I'm a little bit hesitant to make the public ABI dependent on `__cpp_lib_generic_associative_lookup`. If the change is worth it, I'd raise the required C++ standard to 14 (or newer).

But most of the comparisons happen with `Labels` and thus with `std::string`.

Thanks,
Gregor

aeroyorch:
> Hello,
> 
> do you have a benchmark that shows improvements with this change? I'm a little bit hesitant to make the public ABI dependent on `__cpp_lib_generic_associative_lookup`. If the change is worth it, I'd raise the required C++ standard to 14 (or newer).
> 
> But most of the comparisons happen with `Labels` and thus with `std::string`.
> 
> Thanks,
> Gregor

Ok, I'll try to put together a benchmark next week. I don't expect a big improvement inside the library itself, but it could help users who use the `Labels` alias directly, especially when doing lookups with `std::string_view` or string literals.

aeroyorch:
Added a benchmark (`core/benchmarks/labels_bench.cc`), plain `std::map` vs `std::less<>`, looked up with a `std::string_view`. g++ 13, release, medians:

| entries / key len | plain (temp string) | `std::less<>` |
|---|---|---|
| 4 / 8   | 24.5 ns | 17.6 ns |
| 16 / 8  | 30.3 ns | 27.0 ns |
| 4 / 32  | 38.5 ns | 17.1 ns |
| 16 / 32 | 44.1 ns | 24.0 ns |

So roughly 1.1 to 1.4x faster for short keys, and about 2x for keys past the SSO size, where the temporary `std::string` also heap-allocates (1 alloc/lookup vs 0). With a `std::string` key the two are identical within noise, so there is no regression on the path the library itself uses.

No win inside the library then, but a real one for users looking up `Labels` with a `string_view` or literal.

Issue #741, opened 2025-07-22
v0.8.0 throws std::bad_alloc exception


lxh-ob:
Hello Together,

We sometimes get an std::bad_alloc exception when the OS fails to allocate memory.

the coredump backtrace is below.

Is there any ways to prevent process coredump？

Or, would you please catch the exeception and ignore it? It would be better to keep the program running continuously rather than terminating it due to metric issues

The version is v0.8.0. And it seems that other versions will also may trigger the exception.

```
?? ??:0
?? ??:0
?? ??:0
__gnu_cxx::__verbose_terminate_handler() [clone .cold] at vterminate.o:?
__cxxabiv1::__terminate(void (*)()) at ??:?
std::terminate() at ??:?
__cxa_rethrow at ??:?
prometheus::ClientMetric* std::__uninitialized_copy<false>::__uninit_copy<__gnu_cxx::__normal_iterator<prometheus::ClientMetric const*, std::vector<prometheus::ClientMetric, std::allocator<prometheus::ClientMetric> > >, prometheus::ClientMetric*>(__gnu_cxx::__normal_iterator<prometheus::ClientMetric const*, std::vector<prometheus::ClientMetric, std::allocator<prometheus::ClientMetric> > >, __gnu_cxx::__normal_iterator<prometheus::ClientMetric const*, std::vector<prometheus::ClientMetric, std::allocator<prometheus::ClientMetric> > >, prometheus::ClientMetric*) at ??:?
prometheus::ClientMetric* std::uninitialized_copy<__gnu_cxx::__normal_iterator<prometheus::ClientMetric const*, std::vector<prometheus::ClientMetric, std::allocator<prometheus::ClientMetric> > >, prometheus::ClientMetric*>(__gnu_cxx::__normal_iterator<prometheus::ClientMetric const*, std::vector<prometheus::ClientMetric, std::allocator<prometheus::ClientMetric> > >, __gnu_cxx::__normal_iterator<prometheus::ClientMetric const*, std::vector<prometheus::ClientMetric, std::allocator<prometheus::ClientMetric> > >, prometheus::ClientMetric*) at ??:?
[...]

PR #758, opened 2026-08-27
perf: avoid copying MetricFamily in Family::Collect

FujiwaraNaoto:
# perf: avoid copying MetricFamily in Family::Collect

## Summary

`Family<T>::Collect()` built the family in a local variable and returned it via
`return {family};`. A braced-init-list copies its elements, so this deep-copied
the whole `metric` vector (every `ClientMetric` and all of its labels) on every
scrape, temporarily requiring twice the memory of the collected family.

The family is now constructed in place inside the returned vector, so no copy is
made. The redundant `std::move()` around the prvalue returned by
`CollectMetric()` is dropped as well; a prvalue already binds to
`std::vector::push_back(T&&)`.

## Changes

- `core/src/family.cc`: build the returned `std::vector<MetricFamily>` in place
  instead of copy-constructing it from a braced-init-list.
- `core/src/family.cc`: remove the redundant `std::move()` in the collection loop.

## Behavior

No functional change. The returned data is identical; only the number of
allocations and copies performed during a collection is reduced.

`Registry::Collect()` already moves the families out of `Family<T>::Collect()`
via `std::make_move_iterator`, so this removes the last copy on that path.

## Relates to #741

In #741 a `std::bad_alloc` is thrown from the `MetricFamily` copy constructor
while collecting metrics, which is exactly the copy removed here. Cutting the
peak memory required for a collection makes that failure less likely, but this
change does **not** add exception handling, so it does not fully implement the
request in #741 to keep the process alive when an allocation fails.


FujiwaraNaoto:
Could you please review it?

The Bazel CI failure is unrelated to this change — it fails while computing the main repo mapping, before anything is built.

gjasny:
Thanks for fixing!

